PR #66, opened 2019-11-18
Add optional action to run `tfsec` on a repository

triat:
Hello,

I found your tool recently and I totally loved it. As I use terraform regularly, I thought it would be nice to have such a tool as a security checker when someone does a PR on our repos.

This is why I'd like to propose to add a line to reference my action in your README, such that people who want to do the same as me, have already a way to do it and do not struggle to develop it.

Of course, as I'm only reusing your tool, I'm totally open to adapt my version if you think there is something wrong.

Thanks for your time and your tool.

Tom

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ To run:
 docker run --rm -it -v "$(pwd):/workdir" tfsec .
 ```
 
+## Use as an actions
+
+If you want to run tfsec on your repository as an action, you can use the https://github.com/triat/terraform-security-scan which is using `tfsec` the same way you would in local but with an action that you described on your repository.
+
 ## Features
 
 - Checks for sensitive data inclusion across all providers


reedloden:
Could this just be added to this repository itself versus needing a separate repository?

triat:
I think you need a separate repository to create an action and put the releases in the marketplace.

If there is a way, for sure we could merge the projects

liamg:
Thanks @triat! This is a great idea. I agree with @reedloden that it would make sense to keep this in one place if possible, so I'll look at whether we can do that - thanks for your work either way! :tada:

triat:
That would make a lot of sense indeed! Have a look, I will do the same on my side but I think this needs to be separated. 

I'd be happy to find a good way to have this working, let me know if you want to have a quick chat if it helps the integration.

shoekstra:
Hello!

It is possible to keep the action in the same repo but this doesn't seem to be the common trend.  All the actions I've come across thus far have dedicated repos for actions, probably for good reason.

Great project btw!  I came across this while looking for a way to run this via Actions and thought I'd chime in, for now will use @triat's repo ❤️

Vlaaaaaaad:
I also built an action for `tfsec` that posts the output back on the PR with [`reviewdog`](https://github.com/reviewdog/reviewdog): https://github.com/Vlaaaaaaad/temp-tfsec-action

I'm hoping to move it to the reviewdog repo soon, just like https://github.com/reviewdog/action-tflint

triat:
There is also the live report on the PR now (thanks to @shoekstra )

triat:
Hey @liamg, any update on this discussion maybe?

liamg:
Sorry for the delay, and thanks for this, merging now :)

PR #72, opened 2020-01-03
adds KMS auto-rotation check (closes #31)

### LINKED ISSUE #31: Ensure KMS keys are automatically rotated
`aws_kms_key` should enforce that key rotation is enabled.

https://www.terraform.io/docs/providers/aws/r/kms_key.html#enable_key_rotation

bryanlalexander:
- adds AWSNoKMSAutoRotate check (AWS019)
- adds AWSNoKMSAutoRotate test cases

### DIFF
--- a/internal/app/tfsec/aws_no_kms_key_autorotate_test.go
+++ b/internal/app/tfsec/aws_no_kms_key_autorotate_test.go
@@ -12,8 +12,8 @@ func Test_AWSNoKmsKeyAutoRotate(t *testing.T) {
 	var tests = []struct {
 		name                  string
 		source                string
-		mustIncludeResultCode scanner.CheckCode
-		mustExcludeResultCode scanner.CheckCode
+		mustIncludeResultCode scanner.RuleID
+		mustExcludeResultCode scanner.RuleID
 	}{
 		{
 			name:                  "check KMS Key with auto-rotation not set",
--- a/internal/app/tfsec/aws_no_kms_key_autorotate_test.go
+++ b/internal/app/tfsec/aws_no_kms_key_autorotate_test.go
@@ -24,7 +24,7 @@ func Test_AWSNoKmsKeyAutoRotate(t *testing.T) {
 			name: "check KMS Key with auto-rotation disabled",
 			source: `
 resource "aws_kms_key" "kms_key" {
-	enable_key_rotation = "false"
+	enable_key_rotation = false
 }`,
 			mustIncludeResultCode: checks.AWSNoKMSAutoRotate,
 		},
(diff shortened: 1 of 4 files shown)


internal/app/tfsec/checks/aws_no_kms_key_autorotate.go
```diff
@@ -0,0 +1,45 @@
...
+			if keyRotationAttr == nil {
+				return []scanner.Result{
+					check.NewResult(
+						fmt.Sprintf("Resource '%s' does not have KMS Key auto-rotation enabled.", block.Name()),
+						block.Range(),
+					),
+				}
+			}
+
+			if keyRotationAttr.Type() == cty.String && keyRotationAttr.Value().AsString() != "true" {
```
liamg: Minor change but I think this should be a bool type instead of string:
```suggestion
			if keyRotationAttr.Type() == cty.Bool && keyRotationAttr.Value().False() {
```

internal/app/tfsec/aws_no_kms_key_autorotate_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{
+			name:                  "check KMS Key with auto-rotation not set",
+			source:                `resource "aws_kms_key" "kms_key" {}`,
+			mustIncludeResultCode: checks.AWSNoKMSAutoRotate,
+		},
+		{
+			name: "check KMS Key with auto-rotation disabled",
+			source: `
+resource "aws_kms_key" "kms_key" {
+	enable_key_rotation = "false"
```
liamg:
```suggestion
	enable_key_rotation = false
```

internal/app/tfsec/aws_no_kms_key_autorotate_test.go
```diff
@@ -0,0 +1,47 @@
...
+}`,
+			mustIncludeResultCode: checks.AWSNoKMSAutoRotate,
+		},
+		{
+			name: "check KMS Key with auto-rotation enabled",
+			source: `
+resource "aws_kms_key" "kms_key" {
+	enable_key_rotation = "true"
```
liamg:
```suggestion
	enable_key_rotation = true
```

bryanlalexander:
@liamg I fixed those issues, we should be good to go.

off-topic: It seems that adding "helpers" would greatly reduce the level of effort for adding new checks. Would you be interested in reviewing a PR that implements a few of those? (e.g. IsAttrSet("attribute_name") which returns a CheckFunc)

liamg:
@bryanlalexander Awesome, thank you! 🚀 

re: helpers, yep, that sounds very useful, I'd be delighted with a PR 👍

liamg:
Oh I spoke too soon - the signature changed for `NewResult` and `NewResultWithValueAnnotation` to now return a severity. Error being "this is definitely a liability", and warning being "this may be a liability depending on context". Would you mind adding those in, then we can merge? 🚀 🚀 🚀

bryanlalexander:
@liamg We should be good to go now.

liamg:
@bryanlalexander Perfect, thank you! 🚀

liamg:
v0.18.0 is building now with this included.

PR #94, opened 2020-03-16
add soft-fail flag

frenchi:
Hello, 

We'd like to run tfsec in CI without breaking the build. This PR adds the --soft-fail flag to ensure that the exit code always returns `0`. Reports are otherwise unchanged.

Screen Shot 2020-03-16 at 4 57 54 PM


liamg:
Thanks! :rocket:

PR #98, opened 2020-03-27
Added new arguments to exclude some checks globaly

unk1nd0n3:
Hi Liam,

please consider my PR to add new argument to disable/ignore provided checks globally in whole project. Idea is to turn off/skip checks which generate False Positive findings for particular repository

Possible scenarios where this can help:
1. GEN003 finds issues in standard K8S secret manager resources
2. Disable other unused Cloud Providers

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 
 # ignore windows compiled binary
 tfsec.exe
+.vscode/launch.json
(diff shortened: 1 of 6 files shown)


liamg:
Awesome, thanks for the PR :rocket:

Issue #36, opened 2019-11-11
Ensure ECR repository image scans are enabled


reedloden:
`aws_ecr_repository` should enforce that ECR repository image scans are enabled.

https://www.terraform.io/docs/providers/aws/r/ecr_repository.html#image_scanning_configuration

jdheyburn:
Believe this can be closed, given the PR has been merged

PR #101, opened 2020-04-05
Ensure ecr repository image scans are enabled

valll94:
Adding check for ECR image scan enabled or disabled
#36

### DIFF
--- a/internal/app/tfsec/aws_no_ecr_repository_scan_test.go
+++ b/internal/app/tfsec/aws_no_ecr_repository_scan_test.go
@@ -1,0 +1,57 @@
+package tfsec
+
+import (
+	"testing"
+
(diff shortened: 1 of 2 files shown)


internal/app/tfsec/aws_no_ecr_repository_scan_test.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		{
+			name: "check ECR Image Scan disabled",
+			source: `
+resource "aws_ecr_repository" "foo" {
+  name                 = "bar"
+  image_tag_mutability = "MUTABLE"
+
+  image_scanning_configuration {
+    scan_on_push = false
```
liamg, line 44: This test should include `scan_on_push = true` as it ensures there are no false positives.
```suggestion
    scan_on_push = true
```

PR #106, opened 2020-04-28
Bump Go to 1.14

gliptak:


gliptak:
https://travis-ci.org/github/liamg/tfsec/jobs/680604403

```
run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory
```

would you be OK with removing ```vendor/*```?

PR #107, opened 2020-04-28
fix: defend against infinite traversal for local modules

### LINKED ISSUE #52: tfsec hangs indefinitely on MacOS 10.14.6 (Mojave)
I runned test on MacOS from scratch and from release file (github.com)
Test runned more 40 mins and not ended (
Check please
I am using modules also in my TF files (tf 0.12.10)

ttacon:
This fixes #52 . The details on the thread were immensely helpful for identifying the issues here.

There were two separate bugs:

1. We were not correctly setting the filepath for all local module references. The `source` for modules was working only when the path pointed to a sibling or child directory (a common use-case). However, if it referenced a parent directory, this would break the parsing directory traversal logic.
2. As we didn't pass any state to the recursive parsing, this opened up the potential for an infinite loop. A situation that could cause this is the following directory structure:

```
/my-module
    fileA
    fileB
    examples/
        simple/
            main.tf // <--- where source = "../.."
```

This PR addresses the first issue by determining when we've set the wrong path and correcting it. It solves the second issue by introducing a mechanism for tracking state for "known" module paths.

Let me know y'alls thoughts on this - I wasn't super excited to introduce a new concept to the library `pathBreaker` (for circuit breaking), but it was the cleanest, minimal addition I could arrive at.

### DIFF
--- a/internal/app/tfsec/parser/parser.go
+++ b/internal/app/tfsec/parser/parser.go
@@ -352,7 +352,7 @@ func (p pathBreaker) add(path string) {
 	p[path] = struct{}{}
 }
 
-// haveSeen returns a boolean denoting if we've seen the given path before.
+// hasSeen returns a boolean denoting if we've seen the given path before.
 func (p pathBreaker) hasSeen(path string) bool {
 	_, ok := p[path]
 	return ok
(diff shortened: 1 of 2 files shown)


liamg:
Amazing, thank you! :rocket:

PR #114, opened 2020-06-01
Remove extra quote in the help text

jverhoeks:
The --help shows an extra quote
```
      --no-color         Disable colored output (American style!)
      --no-colour        Disable coloured output
  -s, --soft-fail        'Runs checks but suppresses error code
```


liamg:
Thanks :+1:

Issue #85, opened 2020-01-31
Junit or simular output for Jenkins


jverhoeks:
Have the errors in junit format makes it easy to integrate in the jenkins or other ci/cd pipelines

samm-git:
+1, also will help for the circleci test results

MorganGeek:
+1 as well. I'm using checkov and tfsec for my terraform and terragrunt pipelines :-) , checkov provides an option for getting outputs in junit format, it would be awesome to have the same with tfsec !

samm-git:
Ok, it seems to be trivial change, will try to do next days

samm-git:
Done, feel free to test my PR. So far tested with CircleCI only, both enterprise and saas.

MorganGeek:
Thank you @samm-git  :) 🤟

wheelq:
> +1 as well. I'm using checkov and tfsec for my terraform and terragrunt pipelines :-) , checkov provides an option for getting outputs in junit format, it would be awesome to have the same with tfsec !

Are you using tfsec and checkov simultaneously?

MorganGeek:
> > +1 as well. I'm using checkov and tfsec for my terraform and terragrunt pipelines :-) , checkov provides an option for getting outputs in junit format, it would be awesome to have the same with tfsec !
> 
> Are you using tfsec and checkov simultaneously?

Yes :-)

PR #122, opened 2020-06-29
Add number of failures (and tests) to testsuite

k1rd3rf:
Bitbucket requires to know the number of tests and failures in order to pick up the test report.

Related to #85 and #112


liamg:
Thanks :+1:

Issue #113, opened 2020-06-01
tfsec 0.21.0 hangs


TLaborde:
It seems there was a change between version 0.19.0 and 0.21.0 and now tfsec hangs indefinitely.

It works in some sub directories, but not at the root of our project. Could it be something related to folder with dot in the name? I may have more time to dig into the issue in the days to come.

ttacon:
Hi @TLaborde , I'd introduced a change in #107 to help defend against a different infinite loop, and it plausibly is that change which is causing the issue for you. Do you have an example directory structure to reference? I'd be happy to put in some cycles to dig in with you.

TLaborde:
Hi @ttacon 

So i was able to isolate the problem a bit more, and it doesn't hangs indefinitely, but it's just very very slow. It seems to be happening if tfsec is directed to a folder one level higher than the one containing some terraform files. Here is a small script to reproduce it, in powershell,assuming tfsec is in the `PATH`

```Powershell
$mainFile = "terraform {}`r`n"
$mainfile += (1..20 | ForEach-Object { 'module "test{0}" {{source = "./test{0}" }}' -f $_ }) -join "`r`n"
New-Item $env:temp\root\dir1\main.tf -ItemType File -Value $mainFile -Force
measure-command {
    . tfsec $env:temp\root\dir1
} # very fast < 1s
Measure-Command {
    . tfsec $env:temp\root\
} #very slow > 20s
```
When running directly in the folder containing the `main.tf`, it's very fast. If I run it one level higher, it takes about 20 seconds for 20 modules. but it seems that it's growing in O(n2), here is draw showing the time in second, depending on the number of modules: https://plotly.com/~TLaborde/1/

unk1nd0n3:
Hi,

yes. This is real issue that hangs all my pipelines. 

Impacted versions:
0.20.0
0.21.0

haavardw:
I have the same issue, but all versions after 0.12.2 fails for me.

dogfish182:
was just looking at tfsec as well but immediately hit this issue, same structure as what @TLaborde  mentioned, 1 module folder. I thought it had hung and never left it to finish as it was longer than 3-4 minutes which would be a no go for a pipeline.

liamg:
Please could you try this again with `v0.22.0`? Hoping this is fixed now :crossed_fingers:

haavardw:
0.22.0 works for me (still slow though, 1min20sec on my repo).
Thanks!

[...]

PR #124, opened 2020-07-02
Add parse result caching

defsprite:
Introduces a parse result caching quick fix

This should fix the "long parsing issue" #113 that we can observe on a variety of repositories. Looking into this, we noticed that the unfortunate mixture of parsing and building evaluation contexts leads to A LOT of useless parsing under certain real-world conditions. 

In order get a quick fix out, we extend @ttacon's band-aid to also cache the parsing result of `parseModuleBlocks` given their fully resolved path.

Please take a look whether this feasible at all or may break things in an unexpected way. Testing this on a few real-world projects seemed OK so far, but that is only anecdotal evidence.

Thank you for creating tfsec! It is a great idea and a great tool! ❤️

### DIFF
--- a/internal/app/tfsec/parser/parser.go
+++ b/internal/app/tfsec/parser/parser.go
@@ -28,10 +28,16 @@ func New() *Parser {
 	}
 }
 
+type ParseResult struct {
+	Blocks
+	cty.Value
+}
+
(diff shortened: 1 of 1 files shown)


liamg:
This looks good to me, and works well in all cases I've tried it in. Thank you for the contribution!

PR #136, opened 2020-07-09
[GKE Checks] Add additional security checks for k8s

Owen-Cummings:
👋 
Thought I'd add some more GKE security checks. I'll do my best to implement any feedback! If this gets added I can create some wiki pages for these rules to provide more context.

### What checks have been added?
- GCP006: Checks if `node_metadata` in `workload_metadata_config` is set to `EXPOSE` or `UNSPECIFIED`
> setting the flag to EXPOSED or UNSPECIFIED disables metadata concealment.
https://cloud.google.com/kubernetes-engine/docs/how-to/protecting-cluster-metadata#create-concealed

- GCP007: Checks if legacy metadata endpoints have been disabled
> Compute Engine's instance metadata server exposes legacy /0.1/ and /v1beta1/ endpoints, which do not enforce metadata query headers.
https://cloud.google.com/kubernetes-engine/docs/how-to/protecting-cluster-metadata#disable-legacy-apis

- GCP008: Checks for use of risky legacy client authentication methods with static user/pass or client certs

> pre-provisioned x509 certificate or static password were the only available authentication methods, but are now not recommended and disabled by default on new clusters since 1.12+
https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#restrict_authn_methods

- GCP009: Checks whether pod security policy enforcement has been enabled for a cluster. Implementing this requires more than enabling this option, but a wiki page containing a more guidance can be done up.

> By default, Pods in Kubernetes can operate with capabilities beyond what they require. You should constrain the Pod's capabilities to only those required for that workload.
https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#admission_controllers

- GCP010: Checks if shielded GKE nodes have been enabled

> Shielded GKE Nodes provide strong, verifiable node identity and integrity to increase the security of GKE nodes and should be enabled on all GKE clusters.
https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#shielded_nodes

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ there are also checks which are provider agnostic.
 | GCP004  | google   | An outbound firewall rule allows traffic to `/0`.
 | GCP005  | google   | Legacy ABAC permissions are enabled.
 | GCP006  | google   | Node metadata value disables metadata concealment.
+| GCP007  | google   | Legacy metadata endpoints enabled.
 
 ## Running in CI
 
--- a/README.md
+++ b/README.md
@@ -169,6 +169,7 @@ there are also checks which are provider agnostic.
 | GCP006  | google   | Node metadata value disables metadata concealment.
 | GCP007  | google   | Legacy metadata endpoints enabled.
 | GCP008  | google   | Legacy client authentication methods utilized.
+| GCP009  | google   | Pod security policy enforcement not defined.
 
 ## Running in CI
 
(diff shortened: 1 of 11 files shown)


liamg:
@Owen-Cummings Thanks very much for this, everything looks great! I'll add you as a collaborator so you can have wiki permissions. :+1:

Issue #146, opened 2020-08-07
Ensure EFS file systems are encrypted


xlz-jgoutin:
See: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_file_system#encrypted

valll94:
I've added the check and open a pull request. Please accept my apologies  for the first PR (it is closed already), the branch has some mistakes.

PR #151, opened 2020-08-12
Efs encryption

valll94:
#146 
Adding a check about encryption of EFS


liamg:
It looks like your tests are failing in the build here @valll94

owenrumney:
updated in https://github.com/tfsec/tfsec/pull/488 - thanks @valll94  for kicking it off

PR #153, opened 2020-08-13
fix: Update check to TLSv1.2_2019

### LINKED ISSUE #152: Add Support for TLSv1.2_2019
Screen Shot 2020-08-13 at 11 18 39 AM

AWS now has "TLSv1.2_2019" available and uses it as the default. I am receiving the attached Problem notice from tfsec when using the 2019 version instead of 2018.

I think it is safe to say that use of the 2019 version should not be considered "outdated"

mhester-nutrien:
Fixes #152 
Change TLS check from TLSv1.2_2018 to TLSv1.2_2019

### DIFF
--- a/internal/app/tfsec/checks/aws_cloudfront_outdated_protocol.go
+++ b/internal/app/tfsec/checks/aws_cloudfront_outdated_protocol.go
@@ -36,10 +36,10 @@ func init() {
 						scanner.SeverityError,
 					),
 				}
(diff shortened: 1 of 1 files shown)


reedloden (approved):

limakzi (approved):

liamg:
Hi @mhester-nutrien - it looks like we have another PR for the same changes at #160 which includes  a test update so will close this one - thank you though! :+1:

PR #157, opened 2020-09-05
feat: add check for encryption at rest for Elasticsearch domains

ttacon:
This PR adds a few new AWS checks for a few best practices on Elasticsearch domains. I'd love for these to be included, let me know your thoughts @liamg! Specifically, it checks for the these blocks:

```hcl
encrypt_at_rest {
    enabled = true
}
``` 

```hcl
node_to_node_encryption {
    enabled = true
}
```

```hcl
domain_endpoint_options {
    enforce_https = true
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,9 +157,10 @@ there are also checks which are provider agnostic.
 | AWS023  | aws      | ECR repository has image scans disabled
 | AWS024  | aws      | Kinesis stream is unencrypted
 | AWS025  | aws      | API Gateway domain name uses outdated SSL/TLS protocols.
-| AWS031  | aws      | AWS Elasticsearch domain isn't encrypted at rest.
-| AWS032  | aws      | AWS Elasticsearch domain uses plaintext traffic for node to node communication.
-| AWS033  | aws      | AWS Elasticsearch doesn't enforce HTTPS traffic.
+| AWS031  | aws      | Elasticsearch domain isn't encrypted at rest.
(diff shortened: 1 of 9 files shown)


reedloden:
Perhaps also check the TLS protocol version? Ensuring that at least TLS 1.2 is used. Not sure if there's a good way outside of checking to see if TLS 1.0 is selected (and warning on that)?

reedloden:
> Perhaps also check the TLS protocol version? Ensuring that at least TLS 1.2 is used. Not sure if there's a good way outside of checking to see if TLS 1.0 is selected (and warning on that)?

Specifically, DomainEndpointOptions TLSSecurityPolicy should be set to something other than `Policy-Min-TLS-1-0-2019-07` (which is the default).

reedloden:
Fixes https://github.com/liamg/tfsec/issues/34

ttacon:
For sure! I almost wonder if that should be a separate check? In my mind, there are two things we're checking on:

 - Is HTTPS enforced?
 - Is a security cipher suite enabled?

In my mind, in particular as the latter one will evolve over time, I feel that it would be better as another check. I can add one next week to this PR.

reedloden:
Agreed, separate check is fine. Just pointing it out since you were working
on this particular resource.

ttacon:
Absolutely! Adding that check makes a ton of sense.

liamg (approved):
Thank you @ttacon :+1:

PR #158, opened 2020-09-07
feat: add checks for encryption at rest and in transit for Elasticache

ttacon:
This PR adds checks for ensuring that encryption at rest and encryption in transit are set for Elasticache Replication Groups.


liamg (dismissed):
Thanks again @ttacon

liamg (approved):

PR #159, opened 2020-09-09
[GCP Check] Add IAM user grant check

Owen-Cummings:
👋 Adding another GCP check that looks for IAM grants directly to user identities. It searches data types `google_iam_policy` and quite a few resource types related to IAM to accomplish this. All it searches for is a prefix of `user:` within the member/members attribute of each block.

Example TF blocks it would flag:
```
data "google_iam_policy" "test-policy" {
	binding {
                ...
		members = [
			"group:test@example.com",
			"user:test@example.com",
		]
	}
}

resource "google_storage_bucket_iam_binding" "bucket-binding" {
        ...
	members = [
		"user:test@example.com",
		]
}

resource "google_project_iam_member" "project-member" {
        ...
	member = "user:test@example.com"
}
```

If I've forgotten any resources from the check feel free to add them. Otherwise once this gets merged I'll put together a documentation page 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ there are also checks which are provider agnostic.
 | GCP008  | google   | Legacy client authentication methods utilized.
 | GCP009  | google   | Pod security policy enforcement not defined.
 | GCP010  | google   | Shielded GKE nodes not enabled.
+| GCP010  | google   | IAM granted directly to user.
 
 ## Running in CI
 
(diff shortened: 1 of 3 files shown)


liamg (approved):
Thanks @Owen-Cummings :rocket:

MPV:
Just a heads-up that I'm getting warnings like these when running this (from version 0.27.0):

```
$ tfsec
WARNING: skipped GCP011 check due to error(s): value is null
WARNING: skipped GCP011 check due to error(s): value is null
WARNING: skipped GCP011 check due to error(s): value is null
WARNING: skipped GCP011 check due to error(s): value is unknown

No problems detected!

0 potential problems detected:
```

Not sure which of the resources it's warning about in my project, supposedly one of the `*_iam_*` ones. Should say that those have been applied successfully though, so they're valid terraform at least.

PR #424, opened 2020-10-14
Add verbose logging and increase efficiency of evaluation recursion

liamg:


internal/app/tfsec/debug/debug.go
```diff
@@ -0,0 +1,16 @@
...
+	"time"
+)
+
+var Enabled bool
+
+func Log(format string, args ...interface{}) {
+	if !Enabled {
+		return
+	}
+	line := fmt.Sprintf(format, args...)
```
owenrumney, line 14: What do you think about using runtime.Caller to get the calling function?

Might be a useful detail to include in the log line?

owenrumney (approved):

internal/app/tfsec/debug/debug.go
liamg: Yeah, good idea. It'll be worth it as we begin to add debug lines everywhere...

PR #426, opened 2020-10-15
Add doc structs

liamg:
Should have done this manually...
```bash
#!/bin/bash

set -exo pipefail

fix_check () {
	file="$1"
	target=/tmp/target
	cp $file $target
	line=$(grep -n RuleSummary "$file" | head -n1 | awk -F':' '{print $1}')
	prefix=$(cat "$file" | grep 'const' | grep RuleSummary | head -n1 | awk '{print $2}' | sed 's/Summary//g' | sed 's/Description//g')
	cat > /tmp/addition <<- EOF
const ${prefix}Explanation = \`

\`
const ${prefix}BadExample = \`

\`
const ${prefix}GoodExample = \`

\`
EOF

	sed -e "${line}r /tmp/addition" "$target" > /tmp/intermediate

	docline=$(grep -n "Summary:" /tmp/intermediate | head -n1 | awk -F':' '{print $1}')
	cat > /tmp/docs <<- EOF
            Explanation: ${prefix}Explanation,
            BadExample:  ${prefix}BadExample,
            GoodExample: ${prefix}GoodExample,
            Links: []string{},
EOF

	sed -e "${docline}r /tmp/docs" /tmp/intermediate > $file
}

for f in `ls ./internal/app/tfsec/checks/`; do
	echo "Modifying $f..."
	fix_check "./internal/app/tfsec/checks/$f"
done

```

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/tfsec/tfsec/internal/app/tfsec/debug"
+
 	"github.com/tfsec/tfsec/internal/app/tfsec/formatters"
 
 	"github.com/liamg/tml"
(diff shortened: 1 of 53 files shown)


owenrumney (approved):
🤞

PR #431, opened 2020-10-15
Sort checks before running to ensure consistent behaviour

liamg:


internal/app/tfsec/scanner/registration.go
```diff
@@ -25,5 +26,8 @@ func RegisterCheck(check Check) {
 
 // GetRegisteredChecks provides all Checks which have been registered with this package
 func GetRegisteredChecks() []Check {
+	sort.Slice(registeredChecks, func(i, j int) bool {
```
owenrumney, line 29: THIS IS A HANDSOME PIECE OF CODE

owenrumney (approved):

internal/app/tfsec/scanner/registration.go
liamg: Hahaha :100:

PR #448, opened 2020-10-17
Verify example code

liamg:


owenrumney (approved):
Nice. I knew that as I did the....examples they would cause a headache at some point

PR #456, opened 2020-10-21
Extract open cidr checks into util function

particleflux:
Extracted the repeated open CIDR checks into a separate utility function.

I realize this was "_unasked for_" and "_not discussed yet_" according to the contributing guide, though I figured it's the only way to move forward with even more of these checks planned.

Consider this a proposal, and open for discussion.

I'm not sure what to do with the similar checks for *azure*, as in that case, there also seems to be `"*"` possible as an open CIDR.

### DIFF
--- a/internal/app/tfsec/checks/utils.go
+++ b/internal/app/tfsec/checks/utils.go
@@ -2,7 +2,9 @@ package checks
 
 import (
 	"github.com/tfsec/tfsec/internal/app/tfsec/parser"
+	"github.com/tfsec/tfsec/internal/app/tfsec/scanner"
 	"github.com/zclconf/go-cty/cty"
+	"strings"
 )
 
 // isBooleanOrStringTrue returns true if the attribute is a boolean and is
@@ -16,4 +18,31 @@ func isBooleanOrStringTrue(val *parser.Attribute) bool {
 	default:
 		return false
 	}
+}
+
+// isOpenCidr returns true if given attribute is an open CIDR block
+func isOpenCidr(attr *parser.Attribute, provider scanner.RuleProvider) bool {
+	if attr.Value().IsNull() {
+		return false
+	}
+
+	var cidrList []cty.Value
+	if attr.Type() == cty.String {
+		cidrList = []cty.Value{attr.Value()}
+	} else {
+		cidrList = attr.Value().AsValueSlice()
+	}
+
+	for _, cidr := range cidrList {
+		if cidr.Type() != cty.String {
+			continue
+		}
+
+		cidrStr := cidr.AsString()
+		if strings.HasSuffix(cidrStr, "/0") || cidrStr == "*" {
+			return true
+		}
+	}
+
+	return false
 }
--- a/internal/app/tfsec/checks/utils_test.go
+++ b/internal/app/tfsec/checks/utils_test.go
@@ -1,6 +1,7 @@
 package checks
 
 import (
+	"github.com/tfsec/tfsec/internal/app/tfsec/scanner"
 	"testing"
 
 	hcl "github.com/hashicorp/hcl/v2"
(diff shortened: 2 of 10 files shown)


owenrumney:
I'm wondering if we could pass the provider into the check and fan out from there?

Have the isOpenCidr(cidr, provider)

Then switch on the provider to handle azure vs aws approach

What do you think?

particleflux:
The `Provider` field of the `Check` struct? Might be an idea, hadn't realized this is already available to the check func.

It's basically only the innermost condition which is different. Otherwise it's always a `ValueSlice()`, and always checked against `lengthInt()` and `IsNull()`

So something like this should work:

```
cidrStr := cidr.AsString()
switch provider {
case GCPProvider:
case AWSProvider:
    if strings.HasSuffix(cidrStr, "/0") {
        return true
case AzureProvider:
    if strings.HasSuffix(cidrStr, "/0") || cidrStr == "*" {
        return true
}
```

No default case, it'll fallback to `false` anyway.

owenrumney:
Yup, that's what I was thinking might work.

particleflux:
Yes, I will try that.

particleflux:
Extended it to be compatible with Azure and GCP as well.

There is one downside to Azure though: With these changes, it'll loose information on the output. Azure used to show the actual affecting prefix in the error message as well, in case it was a list.

I could possibly re-add that, along with adding it to the other providers as well, by returning 2 values from the `isOpenCidr` function. Not sure if it is really relevant though.

internal/app/tfsec/checks/utils.go
```diff
@@ -17,3 +19,38 @@ func isBooleanOrStringTrue(val *parser.Attribute) bool {
...
+	}
+
+	for _, cidr := range cidrList {
+		if cidr.Type() != cty.String {
+			continue
+		}
+
+		cidrStr := cidr.AsString()
+
+		switch provider {
```
owenrumney, line 43: Hey @particleflux  - apologies for the delay on reviewing this. 

I've just been looking at the documentation and it doesn't seem "*" is valid for AWS or GCP, so there is nothing to be gained with this switch by the look of it. We can get away with

```go
if strings.HasSuffix(cidrStr, "/0" || cidrStr == "*" {
    return true
}
```

and remove the switch altogether?

internal/app/tfsec/checks/utils_test.go
```diff
@@ -53,3 +54,115 @@ func Test_isBooleanOrStringTrue(t *testing.T) {
...
+		},
+		{
+			rawExpr:  `["172.16.32.0/20", "172.16.16.0/20", "172.16.0.0/20", "0.0.0.0/0"]`,
+			provider: scanner.AWSProvider,
+			result:   true,
+		},
+		{
+			rawExpr:  `["*"]`,
+			provider: scanner.AWSProvider,
+			result:   false,
```
owenrumney, line 113: if the `switch` is removed, as per the previous comment, this becomes `true`

internal/app/tfsec/checks/utils_test.go
```diff
@@ -53,3 +54,115 @@ func Test_isBooleanOrStringTrue(t *testing.T) {
...
+		},
+		{
+			rawExpr:  `["10.0.0.0/0"]`,
+			provider: scanner.GCPProvider,
+			result:   true,
+		},
+		{
+			rawExpr:  `["*"]`,
+			provider: scanner.GCPProvider,
+			result:   false,
```
owenrumney, line 152: if the `switch` is removed, as per the previous comment, this becomes `true`

internal/app/tfsec/checks/utils.go
particleflux: Changed, avoids unnecessary complexity :+1: 
Also rebased upon current master to have a clean commit history

owenrumney (approved):

owenrumney:
Thanks @particleflux !

PR #460, opened 2020-10-22
logging does not need when s3 bucket is for logging

hgsgtk:
I try to implement that logging does not need when s3 bucket is for logging.

Resolves https://github.com/tfsec/tfsec/issues/458

I am looking for opinions (ex. better implementation ideas).

It may be better to check whether the bucket is specified by other bucket as logging target, but I didn't immediately know how to realize it.


owenrumney:
Hey @hgsgtk, thanks for the contribution we're sorry it's taken a while to respond. 

I like the rationale of this check, my hesitation is around using the ACL as an indicator of intent. I think in this case it's probably okay.

I think the only alternative is to use a local ignore on the bucket definition which possibly is no more desirable.

owenrumney (approved):

hgsgtk:
Thanks a lot!

PR #461, opened 2020-10-22
Add check skeleton generator

owenrumney:
Add a command to create a check skeleton and its associated test.

takes - provider, short description, summary and required types/labels then generates a new test with the next available number.


liamg (approved):
Nice!

Issue #286, opened 2020-10-07
Ensure AKS cluster has Network Policy configured


owenrumney:
Possible check to target the following resource

 - [ ] azurerm_kubernetes_cluster

owenrumney:
closing, tracking in spreadsheet

owenrumney:
closing, tracking in spreadsheet

PR #472, opened 2020-10-25
[#286] Added check for AKS cluster Network Policy

rdemachkovych:
Ensure AKS cluster has Network Policy configured


liamg (approved):
Thank you!

PR #491, opened 2020-10-30
Ingress acl rules

AMMullan:
Have created 2 rules
049: Check ingress rules that have specific ports and allowed from 0.0.0.0/0
050: Check ingress rules that allow ALL ports and allowed from 0.0.0.0/0

I think for the most part it's likely 049 is more often "OK" whereas 050 should almost never be ignored, that's my logic behind the separation.


owenrumney (approved):
lgtm - thanks @AMMullan

PR #494, opened 2020-11-03
Azure Storage Account HTTPS enabled rule created

cmelo05:
Created rule AZ010 to verify if the flag enable_https_traffic_only is enabled in the Terraform Resource azurerm_storage_account. If this flag is not set, it will be automatically be set to true, so I added a nil check on the attribute.

Documentation:
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account
- https://docs.microsoft.com/en-us/azure/storage/blobs/security-recommendations
- https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview

### DIFF
--- a/internal/app/tfsec/checks/azu010.go
+++ b/internal/app/tfsec/checks/azu010.go
@@ -1,0 +1,63 @@
+package checks
+
+import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


owenrumney:
Thanks for this!

owenrumney (approved):

PR #496, opened 2020-11-05
Add GCP check to ensure that GKE node service accounts are overridden

maaslalani:
👋 

Just adding a check to ensure that `service_account` is overridden in `google_container_cluster` and `google_container_node_pool` inside the `node_config` block.

### Checks Added
* GCP012: Checks if `node_config` in `google_container_cluster` or `google_container_node_pool` has a service account defined which is recommended by Google (see below)

Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#use_least_privilege_sa

> By default, nodes are given the Compute Engine default service account, which you can find by navigating to the IAM section of the Cloud Console. This account has broad access by default, making it useful to wide variety of applications, but it has more permissions than are required to run your Kubernetes Engine cluster. **You should create and use a minimally privileged service account to run your GKE cluster instead of using the Compute Engine default service account.**

### DIFF
--- a/internal/app/tfsec/checks/gcp012.go
+++ b/internal/app/tfsec/checks/gcp012.go
@@ -45,10 +45,19 @@ func init() {
 		RequiredLabels: []string{"google_container_cluster", "google_container_node_pool"},
 		CheckFunc: func(check *scanner.Check, block *parser.Block, _ *scanner.Context) []scanner.Result {
 
-			display_block := block.GetBlock("node_config")
-			service_account := display_block.GetAttribute("service_account")
+			if !block.HasBlock("node_config") {
+				return []scanner.Result{
+					check.NewResult(
+						fmt.Sprintf("Resource '%s' does not define the node config and does not override the default service account. It is recommended to use a minimally privileged service account to run your GKE cluster.", block.FullName()),
+						block.Range(),
+						scanner.SeverityError,
+					),
+				}
+			}
+			displayBlock := block.GetBlock("node_config")
+			serviceAccount := displayBlock.GetAttribute("service_account")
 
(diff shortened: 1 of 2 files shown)


owenrumney:
Thanks for this @maaslalani  - It's prompted me to add a new helper function which is just going through then will finish review

internal/app/tfsec/checks/gcp012.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		Provider:       scanner.GCPProvider,
+		RequiredTypes:  []string{"resource"},
+		RequiredLabels: []string{"google_container_cluster", "google_container_node_pool"},
+		CheckFunc: func(check *scanner.Check, block *parser.Block, _ *scanner.Context) []scanner.Result {
+
+			display_block := block.GetBlock("node_config")
+			service_account := display_block.GetAttribute("service_account")
+
+			if service_account == nil || (service_account.Value().Type() != cty.String) || len(service_account.Value().AsString()) == 0 {
```
owenrumney, line 51:
```suggestion			
			if !block.HasBlock("node_config") {
				return []scanner.Result{
					check.NewResult(
						fmt.Sprintf("Resource '%s' does not define the node config and does not override the default service account. It is recommended to use a minimally privileged service account to run your GKE cluster.", block.FullName()),
						block.Range(),
						scanner.SeverityError,
					),
				}
			}
			displayBlock := block.GetBlock("node_config")
			serviceAccount := displayBlock.GetAttribute("service_account")

			if serviceAccount == nil || serviceAccount.IsEmpty() {
```

owenrumney (approved):

PR #498, opened 2020-11-06
Add attribute is empty

owenrumney:
Add check if the attribute is empty
    
    - supporting tests included
    - covers string, tuple, slice, map, object


internal/app/tfsec/parser/attribute.go
```diff
@@ -181,3 +181,16 @@ func (attr *Attribute) IsTrue() bool {
...
+func (attr *Attribute) IsEmpty() bool {
+	if attr.Value().Type() == cty.String {
+		return len(attr.Value().AsString()) == 0
+	}
+	if attr.Type().IsListType() || attr.Type().IsTupleType() {
+		return len(attr.Value().AsValueSlice()) == 0
+	}
+	if attr.Type().IsMapType() || attr.Type().IsObjectType() {
+		return len(attr.Value().AsValueMap()) == 0
+	}
```
liamg, line 194: Not sure how we should treat numbers - return false for all of them `!= 0`? Not sure if negative numebrs should be counted as `empty` or not :s

liamg (changes requested):

internal/app/tfsec/parser/attribute.go
owenrumney: I'm not sure a number can be empty can it? The default for an int is 0

I considered changing to return an error, but that changes the whole interface

liamg (approved):

PR #509, opened 2020-11-12
Don't raise check if `node_config` is empty on `google_container_cluster`

maaslalani:
Looking at the [docs for `google_container_cluster` and `node_config`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#node_config), it says the `node_config` block shouldn't be used together with `google_container_node_pool`.  So, we probably don’t need to flag this check if the `node_config` is not defined in the `google_container_cluster` block since its likely they are not using the default node pool.

### DIFF
--- a/internal/app/tfsec/checks/gcp012.go
+++ b/internal/app/tfsec/checks/gcp012.go
@@ -2,6 +2,7 @@ package checks
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/tfsec/tfsec/internal/app/tfsec/parser"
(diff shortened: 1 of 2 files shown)


owenrumney:
This comment comes with the caveat I have zero GCP -

would this check be better to include a verification that `remove_default_node_pool` is `true` and there is no `node_config` rather than assuming it isn't there because they've also got a `google_container_node_pool`?

maaslalani:
Ahh that is a great suggestion. That would definitely make this check logic better! Let me add that :)

owenrumney (approved):
Like it! Thanks very much for contributing!

PR #510, opened 2020-11-14
Added capital case format output for CSV and JSON

NaorPenso:
For easier user experience and alignment with the README.md (stating that CSV and JSON are supported in capital case).


owenrumney:
Thanks for contribution @NaorPenso  - I absolutely see the logic for allowing uppercase and coming in line with the documentation.

I think this is better achieved by changing line 155 from 

```golang
switch format {
```

to 

```golang
switch strings.ToLower(format) {
```

then we don't need the duplication in the switch block and can still use JSON or json.

What do you think?

NaorPenso:
Agreed, will definitely make the code more tidy and consistant, i'll push a new pull request with the amended changed.

Thanks!
Naor

PR #511, opened 2020-11-14
Set format output to support uppercase & lowercase 

NaorPenso:
Formatter output set to lowercase, supporting uppercase and lowercase output types (CSV, JSON, csv, json etc.)


owenrumney (approved):
Looks good. Thanks again for contributing

PR #513, opened 2020-11-16
Support related link output for custom checks

Owen-Cummings:
👋 Here's a quick PR that updates check output to account for custom checks. This changes the output based on provider so custom checks point to relatedLinks instead of `https://tfsec.dev/docs/custom/CUS001`.

It would be worthwhile to make sure you're happy with how the output is handled for the various other formats (ie: json, csv, checkstyle, junit, sarif). Feel free to tweak this! I just thought I'd put a PR in for this quick fix so custom checks aren't sent to a 404 😄


owenrumney (approved):
Thanks @Owen-Cummings  - are you guys now making use of custom checks?

PR #520, opened 2020-11-17
build statically linked binaries

pschulten:
fixes https://github.com/tfsec/tfsec/issues/519


owenrumney (approved):

owenrumney:
thanks @pschulten - seems like this will be most useful :+1:

PR #526, opened 2020-11-24
Additional custom checks for `notContains` and `requiresPresence`

felipesere:
Add two actions for custom checks:

* `notContains`, which is useful for policy documents on AWS to exclude certain patterns such as `kms:*`.
* `requiresPresence` that requires that the resource mentioned in the `matchSepc` is also present


owenrumney (approved):
this is great, thanks

owenrumney:
thanks @felipesere  - I really like this and thanks for the time taken to make sure the docs are in order.

PR #527, opened 2020-11-24
Support yaml custom check config files

### LINKED ISSUE #523: Support custom check definition with YAML or other languages
It would be a nice feature if there was an option to use various languages such as YAML when writing custom checks. The main selling point for YAML is how human readable it is in addition to supporting comments. Working with json has mainly been an issue because it doesn't support comments. This would be beneficial as currently the only way to add context within the check file is in the description field.

Having a more robust tfsec related file parser could be leveraged down the road when adding config files and other features. It's definitely a nice to have but not a need to have. 

An update to `custom_rules.go` that supports various languages. Making it able to handle more than json would be a good way to expand possible use cases similar to how tfsec supports different forms of output.

Write checks in json. 

**Additional context**
Sample of what a rule in YAML would look like
```yaml
---
checks:
- code: CUS001
  description: Does things
  requiredTypes:
  - resource
  - data
  requiredLabels:
  - cool_resource
  - cool_data
  severity: WARNING
  matchSpec:
    name: neat_block
    action: isPresent
    subMatch:
      name: nice_value
      action: equals
      value: good
  errorMessage: your nice_value is not good
  relatedLinks:
  - very-cool-website.co.uk
```

owenrumney:
resolves #523 

fyi @Owen-Cummings


internal/app/tfsec/custom/custom_check.go
```diff
@@ -80,23 +80,23 @@ const RequiresPresence CheckAction = "requiresPresence"
...
-	IgnoreUndefined bool        `json:"ignoreUndefined,omitempty"`
+	Name            string      `json:"name,omitempty" yaml:"name,omitempty"`
+	MatchValue      interface{} `json:"value,omitempty" yaml:"value,omitempty"`
+	Action          CheckAction `json:"action,omitempty" yaml:"action,omitempty"`
+	SubMatch        *MatchSpec  `json:"subMatch,omitempty" yaml:"subMatch,omitempty"`
+	IgnoreUndefined bool        `json:"ignoreUndefined,omitempty" yaml:"ignoreUndefined,omitempty"`
 }
 
-// CustomCheck specifies the check definition represented in json
+// CustomCheck specifies the check definition represented in yaml
```
liamg, line 90: nitpicking here but CustomCheck should be Check in the godoc?
```suggestion
//Check specifies the check definition represented in json/yaml
```

liamg (changes requested):

liamg (approved):

PR #532, opened 2020-12-03
Also load exclude list from config and merge with flags

felipesere:
Also fixes a small bug where config files with the suffix `yml` where not being loaded correctly.

### DIFF
--- a/docs-website/_docs/config.md
+++ b/docs-website/_docs/config.md
@@ -14,18 +14,40 @@ The structure of the config file can be either `json` or `yaml` and is passed us
 ```json
 {
   "severity_overrides": {
-    "CUS002": "ERROR",
+    "CUSTOM001": "ERROR",
     "AWS025": "WARNING"
   }
 }
 ``` 
 
 or 
+
 ```yaml
 ---
 severity_overrides:
-  CUS002: ERROR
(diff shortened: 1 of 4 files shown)


docs-website/_docs/config.md
```diff
@@ -14,20 +14,42 @@ The structure of the config file can be either `json` or `yaml` and is passed us
 ```json
 {
   "severity_overrides": {
-    "CUS002": "ERROR",
+    "CUSTOM001": "ERROR",
```
owenrumney, line 17: the code for custom checks is arbitrary but the choice to document and guide towards `CUS0XX` is deliberate to maintain consistency across the custom and builtin checks when reviewing output reports. I don't think anything is gained by updating this to CUSTOM

internal/app/tfsec/config/config.go
```diff
@@ -33,8 +34,7 @@ func LoadConfig(configFilePath string) (*Config, error) {
 		if err != nil {
 			return nil, err
 		}
-	case ".yml":
-	case ".yaml":
+	case ".yaml", ".yml":
```
owenrumney, line 37: good spot

owenrumney (approved):
lgtm, thanks for this!

PR #536, opened 2020-12-14
AWS KMS key rotation not available for SIGN_VERIFY keys

marcinwyszynski:
This should allow folks to avoid ignoring cases they can do nothing about.

### DIFF
--- a/internal/app/tfsec/checks/aws019.go
+++ b/internal/app/tfsec/checks/aws019.go
@@ -43,7 +43,7 @@ func init() {
 		CheckFunc: func(check *scanner.Check, block *parser.Block, _ *scanner.Context) []scanner.Result {
 			keyUsageAttr := block.GetAttribute("key_usage")
 
-			if keyUsageAttr != nil && keyUsageAttr.Value().AsString() == "SIGN_VERIFY" {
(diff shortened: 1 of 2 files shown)


internal/app/tfsec/checks/aws019.go
```diff
@@ -41,13 +41,14 @@ func init() {
 		RequiredTypes:  []string{"resource"},
 		RequiredLabels: []string{"aws_kms_key"},
 		CheckFunc: func(check *scanner.Check, block *parser.Block, _ *scanner.Context) []scanner.Result {
-			keyRotationAttr := block.GetAttribute("enable_key_rotation")
 			keyUsageAttr := block.GetAttribute("key_usage")
 
-			if keyUsageAttr.Value().AsString() == "SIGN_VERIFY" {
+			if keyUsageAttr != nil && keyUsageAttr.Value().AsString() == "SIGN_VERIFY" {
```
owenrumney, line 46:
```suggestion
			if keyUsageAttr != nil && keyUsageAttr.Equals("SIGN_VERIFY") {
```

there are a number of helper functions that aim to keep these checks concise

owenrumney (approved):

PR #540, opened 2020-12-15
Fixing false positives with rules related to GCP

fallard84:
I tried using tfsec for our GCP projects and had many false positives. This is addressing some of them:

- Adding `google_secret_manager_secret` to whitelist of GEN003. This resource will not contain secrets but the `secret_id` is likely to be named with a sensitive attribute (e.g. secret or token) since this is a secret manager after all!
- GCP011 was failing with error when the identities in the members attribute were unknowns. 
- GCP012 was creating false positive when the `service_account` was using a reference expression.


owenrumney (approved):
looks goodm thanks

PR #551, opened 2021-01-11
Matchspec contains autoscaling grp support

sebay:
Matchspec contains autoscaling grp support


owenrumney (approved):
Looks good, thanks :+1:

PR #553, opened 2021-01-11
Add the description to the JUnit output.

felipesere:
Our users are not as familiar with the RuleIDs
and the description tends to contain a good hint about
what to do next.


owenrumney (approved):
looks good, makes sense. Thanks :+1:

PR #572, opened 2021-02-03
Add default config file

zachwhaley:
tfsec can already load a config file if it explicitly sets one in the `-config-file` option and that file exists, however in our automation we have a common function to run tfsec on any arbitrary code base and not all of these are guaranteed to have a config file in their repo.

I would like tfsec to add a default config file, e.g. `.tfsec/config.yml`, that if it exists is loaded and used as _the_ config file.  This file would be superseded by setting the option `-config-file`.

I considered adding a `tfsec.yml` file to every repo that tfsec is run on, but thought it would be easier to just add this to tfsec instead 😄 

I also decided to reorganize the config docs in a way I thought made more sense.  Feel free to request I revert those changes or rewrite them as you see fit.


owenrumney:
Yeah. I like this - and I can see the value it adds.

Thanks!

owenrumney (approved):

owenrumney:
Building as v0.37.2 - https://github.com/tfsec/tfsec/releases/tag/v0.37.2

PR #573, opened 2021-02-03
Fix issue when loading a bad default config file 

zachwhaley:
If the default config file is badly formed tfsec will not fail, instead it will continue to dereference a nil tfsecConfig pointer and crash.

To avoid this, first check if the default config file exists, if not we'll continue on happily; otherwise load the config and fail if there are any errors.

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -96,32 +96,14 @@ var rootCmd = &cobra.Command{
 		tfsecDir := fmt.Sprintf("%s/.tfsec", dir)
 
 		if len(configFile) > 0 {
-			debug.Log("loading config file...")
-			tfsecConfig, err = config.LoadConfig(configFile)
-			if err != nil {
-				fmt.Fprint(os.Stderr, fmt.Sprintf("Failed to load the config file. %s", err))
-				os.Exit(1)
-			}
-			debug.Log("loaded config file %s", configFile)
+			tfsecConfig = loadConfigFile(configFile)
 		} else {
 			jsonConfigFile := fmt.Sprintf("%s/%s", tfsecDir, "config.json")
 			yamlConfigFile := fmt.Sprintf("%s/%s", tfsecDir, "config.yml")
 			if _, err = os.Stat(jsonConfigFile); err == nil {
-				debug.Log("loading config file...")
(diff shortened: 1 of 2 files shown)


zachwhaley:
Very sorry.  Just found this bug while playing around.

Thanks!

cmd/tfsec/main.go
```diff
@@ -96,19 +96,14 @@ var rootCmd = &cobra.Command{
...
 			yamlConfigFile := fmt.Sprintf("%s/%s", tfsecDir, "config.yml")
-			if tfsecConfig, err = config.LoadConfig(jsonConfigFile); err == nil {
-				debug.Log("loaded config file %s", jsonConfigFile)
-			} else if tfsecConfig, err = config.LoadConfig(yamlConfigFile); err == nil {
-				debug.Log("loaded config file %s", yamlConfigFile)
+			if _, err = os.Stat(jsonConfigFile); err == nil {
+				tfsecConfig = loadConfigFile(jsonConfigFile)
+			} else if _, err = os.Stat(yamlConfigFile); err == nil {
+				tfsecConfig = loadConfigFile(yamlConfigFile)
 			}
```
owenrumney, line 107: this seems to have introduced a but. You need to ensure that `tfsecConfig` doesn't come out of this nil. It's used later and is expected to be a skeleton of a Config

owenrumney, line 107:
```suggestion
			} else {
			       tfsecConfig = &config.Config{}
			}
```

zachwhaley: I believe this takes care of that since execution won't enter this block unless a json or yaml config file exists and then `loadConfigFile` will not return a `nil` Config since it checks for `err` and fails before then.  But I could have missed something.

owenrumney: I was going off the sarif check failing suggested that by the time tfsecConfig was used it had become nil.

owenrumney: ahhhhh, its possible that the sarif check is being crap and checking  against master. but I'm sure it does PR scanning

zachwhaley: I see your edit 👍🏻

owenrumney:
Think we're good! Thanks again!

PR #580, opened 2021-02-08
Allows for multiple submatches that are logically ANDed

felipesere:


owenrumney:
Hey @felipesere  - does this mean you're actively using the custom checks? I think this is a great addition :+1:

felipesere:
We live and breath custom checks :) We have about 10 or so.
I can probably share them and possibly turn them into specific AWS checks? They cover most stuff we pick up with ScoutSuite and SecurityHubs `AWS Foundational Best Practices Benchmark`.

PR #585, opened 2021-02-09
Make contains for strings and lists of strings case insensitive

felipesere:


owenrumney:
We have EqualityOptions https://github.com/tfsec/tfsec/blob/19bf7bed6b1ae8a32f0bda256d91f17ece7efc9f/internal/app/tfsec/parser/attribute.go#L112

I wonder if we could change this to `MatchOptions` and use the same approach for contains? What do you think?

felipesere:
Oh nice, yeah I'll rework this after #587 lands.

owenrumney (approved):
lgtm

PR #587, opened 2021-02-10
Add support for "and" and "or" predicates

alloveras:
## Context
Support for `AND` predicates was added as part of https://github.com/tfsec/tfsec/pull/580. 

## Intent
To re-work the current implementation for composite checks.

## Problem
The major downsides we found with the current implementation are:
- Some custom checks may benefit from an `or` operator being available.
- The implicit `and` behavior of `subMatches` is not apparent from the check definition files. In consequence, non-experienced users will struggle to understand/write checks without having to refer to the official documentation or rely on their teams' tribal knowledge.
- The `subMatch` / `subMatches` operator is a bit overloaded with too many responsibilities:
    -  Provide the ability to “look into” a nested block. 
    -  Provide support for `and`-based composite boolean expressions.

## Proposed Solution
To adopt a new implementation for composite checks that supports both `and` and `or` operators, it is explicitly visible from the check definition files, and do not overload the `subMatch` operator keeping the `chekFunctions` as atomic as possible.

As you can see from the check_all_statements_test.go file, it will still be possible to provide the same functionality that `subMatches` had provided, while allowing for general checkFunctions to be combined. 

(cc: @muplim)

### DIFF
--- a/docs-website/docs/config/custom_checks.md
+++ b/docs-website/docs/config/custom_checks.md
@@ -98,6 +98,7 @@ The `MatchSpec` is the what will define the check itself - this is fairly basic 
 | value | In cases where a value is required, the value to look for |
 | ignoreUndefined | If the attribute is undefined, ignore and pass the check |
 |subMatch | A sub MatchSpec block for nested checking - think looking for `enabled` value in a `logging` block |
+|subMatches | An array of sub MatchSpec block for nested checking. All sub MatchSpecs must match. |
 
 #### Check Actions
 There are a number of `CheckActions` available which should allow you to quickly put together most checks.
--- a/internal/app/tfsec/custom/check_all_statements_test.go
+++ b/internal/app/tfsec/custom/check_all_statements_test.go
@@ -44,7 +44,7 @@ func init() {
 			"action": "isPresent",
 			"subMatch": {
 				"action": "and",
-				"childMatchSpec": [{
+				"predicateMatchSpec": [{
 					"name": "actions",
 					"action": "notContains",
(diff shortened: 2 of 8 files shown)


owenrumney:
This makes sense, thank @alloveras - Could you update the documentation to reflect this change as well please - https://github.com/tfsec/tfsec/blob/master/docs-website/docs/config/custom_checks.md

@felipesere  - this will be relevant to you as it would be a breaking change to your new addition. I think given its early stages I trust that won't impact you too heavily

felipesere:
This looks really good! I'd be happy for this!

owenrumney (changes requested):
All good - just the doc to update - https://github.com/tfsec/tfsec/blob/master/docs-website/docs/config/custom_checks.md

alloveras:
> All good - just the doc to update - https://github.com/tfsec/tfsec/blob/master/docs-website/docs/config/custom_checks.md

Sure, no problem. I will get to that as soon as I have some spare minutes and send it back for a final review.

alloveras:
Hey @owenrumney,
I added the changes to the documentation you requested.

Do you mind having another look?

Thanks!

owenrumney:
This is great, thanks @alloveras  - will cut it as v0.38.1

alloveras:
No worries @ owenrumney 👍 

Actually, we have another PR incoming (next couple of minutes) that I'd like to consider for 0.38.1. 

Would you mind taking a look before cutting the release?

owenrumney:
I already cut one, but happy to do the next once you've raised your PR

Will keep an eye out

PR #597, opened 2021-02-15
Switch to using scratch for docker base image

pujan14:
Signed-off-by: Pujan Shah <pujan14@gmail.com>

Switch to using scratch for docker base image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,13 +11,10 @@ RUN go build \
   -mod=vendor \
   ./cmd/tfsec
 
+###
+FROM scratch
 
-FROM alpine
-
-# use a non-privileged user
-USER nobody
-
-# work somewhere where we can write
+# Copy tfsec from build container
 COPY --from=build-env /src/tfsec /usr/bin/tfsec
 
 # set the default entrypoint -- when this container is run, use this command


owenrumney:
Hey @pujan14, thanks for the PR. I’m not sure this is something we want to do. I think the benefits of alpine with a nobody user offer more flexibility for downstream users.

thoughts @liamg ?

pujan14:
Hi @owenrumney,
I think in this case docker is used only as packaging/distribution format. There are no dependencies inside container and usually downstream users will not need to change anything inside container so I don't see any advantage of using alpine.

owenrumney:
@pujan14 we actually use it in a downstream suite where we mount override the entry point to suit our needs. That would be much more difficult as a scratch image.

that said, this might be a good opportunity to update travis to write multiple flavours of the docker image. Will look at that in the morning

owenrumney:
Hey @pujan14  - I've added `tfsec/tfsec-alpine` and `tfsec/tfsec-scratch` to the build. We discussed and we don't want to change the existing one because it may impact more downstream users than just us.

Thanks for the PR, it prompted us to do this

PR #599, opened 2021-02-16
chore: add dependabot to manage dependencies

chenrui333:
👋  I have introduced dependabot into many corporate and OSS projects (I think it would be a great tool to help managing the dependencies for this great project).


.github/dependabot.yml
```diff
@@ -0,0 +1,24 @@
...
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: weekly
+
+  # Maintain dependencies for dockerfile
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
owenrumney, line 24:
```suggestion
      interval: weekly
```

Looks good, does this need quotes?

chenrui333: it is yaml config, either way is fine :)

chenrui333: in the example, you can see it is quoted
https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#scheduleday

owenrumney: my query was more around the inconsistency. we have 2 unquoted and 1 not

chenrui333: I got you, let me fix it. :)

chenrui333: all good now.

owenrumney (approved):

owenrumney:
all good thanks for this

PR #610, opened 2021-02-19
Fix AWS051 `value is null` warning

particleflux:
When `aws_rds_cluster` attribute `kms_key_id` was explicitly set to
`null`, tfsec did not catch the disabled encryption and issued a
warning:

> WARNING: skipped AWS051 check due to error(s): value is null

The added test case reproduces this issue.

**Note**: The test case does not _catch_ the issue though, as the test execution
logic seems to differ from the real scanning logic, in that it doesn't
try to format the result.

### DIFF
--- a/internal/app/tfsec/checks/aws051.go
+++ b/internal/app/tfsec/checks/aws051.go
@@ -47,7 +47,7 @@ func init() {
 			kmsKeyIdAttr := block.GetAttribute("kms_key_id")
 			storageEncryptedattr := block.GetAttribute("storage_encrypted")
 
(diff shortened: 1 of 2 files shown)


particleflux:
The actual error source is probably here: https://github.com/tfsec/tfsec/blob/6155d330bbb3b1c501fd431cc8242b86516d59f3/internal/app/tfsec/checks/aws051.go#L69-L74

Where `check.NewResultWithValueAnnotation` complains about `kmsKeyIdAttr` being null

owenrumney (approved):

Issue #623, opened 2021-02-23
Option to include passed checks in result output


Malagari:
This library is quickly becoming the recommended option for automatic IaC policy validation within CI pipelines.  Most of these CI applications include a way of publishing and/or displaying test results - unit tests, integration tests, and now security scans.

The current result output only shows failed tests.  There are two problems I see with this --

1. If no checks applied, the user is given a false sense of security that their package is good to go.  Imagine this in the case of unit tests - the lack of tests is not the same as "all tests passed".

2. For users with a significant number of custom checks, it's hard to identify which ones applied to a particular Terraform module.

Ideally, there should be a way to output all of the checks that applied, as well as the result status - Passed or Failed.  This could be enabled via a command line argument (`--extended-output`).  This would allow CI pipelines to automatically publish all applicable checks, or fail builds that did not have a proper number of checks applied (i.e. must have at least one check).

Malagari:
Opened #627 to solve for this. Waiting on feedback from the maintainers.

owenrumney:
Hey @Malagari - commented on the original PR https://github.com/tfsec/tfsec/pull/624#issuecomment-785376642

Malagari:
Thanks @owenrumney, I'll take a look at the snippet you referenced and see if I can make something work there.  Admittedly this was a first pass at the problem and it started to touch more places than I would have liked.  Now that I'm more familiar with the repo, I'll see if there's a way I can simplify the solution.  I'm going to kill #627 in the meantime to not clutter the PR list.

owenrumney:
That's cool @Malagari  - we're always welcoming off contributions and I'm more than happy to work through this one with you. See what you come up with and I'll sleep on it and see what I wake up with. it may not be the option I think it is but worth a look!

Malagari:
This seems to be _way_ cleaner than the first pass I took at it.  I still need to update the default formatter, but the output is generating correctly with less intrusion in to the rest of the codebase.  The Scanner object could also eventually take a set of options rather than passing in individual parameters from the startup.

https://github.com/Malagari/tfsec/pull/2/files

owenrumney:
Hey @Malagari  - sorry for the delay in merging, its in now

PR #624, opened 2021-02-23
Add --include-passed flag to add passed checks to result output.

Malagari:
References #623 

This PR adds a boolean argument to `--include-passed` checks in the result output.  This adds a boolean property on the `Result` object that is set to `true`/`false` depending on whether the check has passed or not.  If the command line argument is set, the passed results will display in the result output.

Additionally, I've updated the default formatter to reflect proper colors (green) for passing checks.  I also added some additional metrics (Total, Passed, Failed Checks) to the statistics output when the flag is passed.  This **does** respect the existing `--concise-output` flag.

Example:

image

This is my first time tinkering with Go, so feel free to nitpick as you see fit. I'm open to feedback.

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -2,11 +2,12 @@ package main
 
 import (
 	"fmt"
-	"github.com/tfsec/tfsec/internal/app/tfsec/config"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
+
+	"github.com/tfsec/tfsec/internal/app/tfsec/config"
 
 	"github.com/tfsec/tfsec/internal/app/tfsec/custom"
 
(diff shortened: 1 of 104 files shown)


Malagari:
Noticed the failing build.  Will re-open once I address those test cases.

owenrumney:
Hey @Malagari  - I see where you're going with this, and its not the first request of its type. I had a look through the PR (I appreciate it's been closed for repair for now) and it felt quite intrusive and potentially adds a memory overhead for an optional requirement.

I wonder if it might be easier/cleaner/(resource)cheaper to target this here -> https://github.com/tfsec/tfsec/blob/786778b1bffa13b5e80359671aea0517ebfba70f/internal/app/tfsec/scanner/scanner.go#L53

if we can pass the argument down to the scanner it could optionally add something here if required - might be worth a look

PR #627, opened 2021-02-24
Add argument to include passed checks in output

Malagari:
**This is my first time tinkering with Go, so feel free to nitpick as you see fit. I'm open to feedback and people modifying this branch.**

References #623

This PR adds a boolean argument to --include-passed checks in the result output.  This adds a boolean property on the `Result` object that is set to `true`/`false` depending on whether the check has passed or not.  If the command line argument is set, the passed results will display in the result output.

Additionally, I've updated the default formatter to reflect the proper colors (green) for passing checks.  I also added some additional metrics (Total, Passed, Failed Checks) to the statistics output when the flag is passed.  This does respect the existing `--concise-output` flag.

Other notes:
- Ensured that the exit code is respected when all tests pass and the `--include-passed` flag is set.
- Updated the readme to include this flag.
- Checks now require returning a result with the `Passed` property set to true.  I've gone though and updated all of the existing checks, as well as the custom check provider.  There may be a better way to do this.

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,8 @@ American friends).
 You can output tfsec results as JSON, CSV, Checkstyle, Sarif, JUnit or just plain old human readable format. Use the `--format` flag
 to specify your desired format.
 
+The `--include-passed` flag will return all checks - passing or failing - that applied to the Terraform files scanned.
+
 ## Github Security Alerts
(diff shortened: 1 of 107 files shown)


Malagari:
Closing until Owen's feedback can be incorporated.

PR #629, opened 2021-02-25
Fixes nested module loading issues

muplim:
### Problem
Currently, when metadata for modules do not exist, the module loader is unable to find and load modules that may exist in outer layers of the directory. To visualise exactly what I mean: 

```
.
├── foo
    ├── bar
        ├── entrypoint
├── source_module
```
Given the above entrypoint, the source for `source_module` would be `../../../source_module`, the existing code incorrectly constructs the filepath to the module.

### Steps to Reproduce
Here's a sample directory tree to reproduce the issue: 
[tfsec-error.zip](https://github.com/tfsec/tfsec/files/6040951/tfsec-error.zip)

If you run tfsec from `foo/bar/entrypoint` without running `terraform init` first, you will see that the filepath mentioned in the warnings are incorrect. 

### Proposed Solution 
We've added a function to reconstruct the filepath based on the current module path and the provided source path. 

(cc: @alloveras )


alloveras:
Hey @owenrumney,
is there any chance you could take a look at this one? 

We are currently having massive log spam in our `tfsec` logs. After a bit of debugging, we were able to narrow it down to a regression introduced in the `0.38.2` release which has been carried forward until the current latest release `0.39.4`.

Our best-effort dive deep lead us to the proposed fix but, perhaps there is something more elegant we can do to actually correct the problem.

Thanks in advance!

(cc: @muplim)

owenrumney (approved):

PR #630, opened 2021-02-25
Add support for negating predicateMatchSpec

muplim:
### Problem
Currently, `matchSpec` actions are defined in such a way that you must create new `matchSpec` actions everytime you wish to negate a particular action. 

### Proposed Solution 
By introducing a `not` action, we can now support negating all existing actions (which may result in duplicates), but this makes adding new actions in the future much easier and also will simplify some of the custom check processing code. 

### Follow Up
For backwards compatibility, we have not removed existing actions that may be duplicated from the introduction of being able to use `not`, however if you would like, we can introduce a follow-up PR to add deprecation notices for actions that overlap with the introduced `not` functionality. 

(cc: @alloveras )


owenrumney (approved):
I really like this thanks. I'm a little embarassed I didn't add it in the first place!

owenrumney (approved):

PR #635, opened 2021-02-26
Add argument to include passed checks in the output

Malagari:
References #623 

Based on feedback to the prior PR (#624), I've made this one less intrusive to the codebase. 

This PR adds an argument (`--include-passed`) that will include all checks that passed against the supplied Terraform.

I've added a `Passed` property on to the `Result` class so that it can be output via the different formatters.  This defaults to `false`.  Since the scanner already does the work of comparing the HCL blocks and checks to find the relevant ones, and then runs the actual check function against the block, I simply had to catch checks that _didn't fail_ their checkFunc (i.e. did not return a `result` object) if the flag was set.

By making the change here, we don't need to have additional conditional logic in each of the formatters or checks themselves to include/exclude results.  The checks will either be in the resulting range or not, depending on the flag.

**Other Notes**
- Followed the `FormatterOption` pattern to include options that can be passed in to the scanner.
- Updated the default and text formatters to change up some pretty-print when displaying passing checks.
- Updated the JUnit formatter to have the accurate count of Tests vs. Failures.
- Updated the CSV formatter to include a `passed` column.
- Returns the appropriate exit code based on the existing logic, even when there are passed results.

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -165,7 +165,7 @@ var rootCmd = &cobra.Command{
 		}
 
 		debug.Log("Starting scanner...")
-		results := scanner.New().ScanWithOptions(blocks, mergeWithoutDuplicates(excludedChecksList, tfsecConfig.ExcludedChecks), includePassed)
+		results := scanner.New().Scan(blocks, mergeWithoutDuplicates(excludedChecksList, tfsecConfig.ExcludedChecks), getScannerOptions()...)
 		results = updateResultSeverity(results)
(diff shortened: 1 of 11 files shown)


Malagari:
**Note:** I have two `TODO` sections in the Sarif and checkstyle formatters.  I'm unfamiliar with these output styles and could use some guidance on whether the changes they need are simply adding the `Passed` property similar to the CSV output, or if they require actual schema changes like the JUnit XML.

owenrumney (approved):

PR #651, opened 2021-03-17
Add check for tfvars files and warn if not being used.

### LINKED ISSUE #641: Automatically use file terraform.tfvars when present
Since terraform automatically load variables from terraform.tfvars file, tfsec needs to do the same.

owenrumney:
Closes #641, provide a warning when a tfvars is available but not being used by tfsec

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -157,6 +157,8 @@ var rootCmd = &cobra.Command{
 				fmt.Println(err)
 				os.Exit(1)
 			}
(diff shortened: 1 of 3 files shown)


liamg (approved):

diogomaul:
Hi @owenrumney, I'm 6 months into Terraform and based on what I've read, it seems the correct parameter to check would be -var-file, not --tf-vars. I tried to make TF work using --tf-vars, but it doesn't recognize it. 

-var-file

Should I open a new Issue to discuss this?

PR #657, opened 2021-03-22
Improving sensitive attributes detection

fproulx-boostsecurity:
Sensitive attribute detection is quite rough at the moment. It easily gets false positives with things like `bigcorp_eks_aux__externalsecrets_bingo_version` , where it's a version of an externalsecrets , nothing wrong here....


owenrumney (approved):
looks good, thanks

owenrumney (approved):

owenrumney:
@fproulx-boostsecurity  - looks like it needs rebasing then we can merge, thanks

PR #659, opened 2021-03-23
add basic support for aliases in ignore

### LINKED ISSUE #91: Add more user-friendly aliases to rules
It'd be nice to be able to say, e.g. `tfsec:ignore:aws-plain-http`  instead of `tfsec:ignore:AWS004`. That way, we can easily see what's being ignored in files.

michal-franc:
closes #91

### Add Support for aliases in `tfsec:ignore`

Example:
Will ignore `AWS018` using `alias` instead of `tfsec:ignore:AWS018` 
```
resource "aws_security_group" "http" {
  name        = "http"

  //tfsec:ignore:aws-security-group-description-missing

  ingress {
    description = "HTTP from VPC"
    from_port   = 80
    to_port     = 80
    protocol    = "tcp"
    cidr_blocks = [aws_vpc.main.cidr_block]
  }
}
```

- Supports backward compatibility, checks without aliases are ignored. To do that: refactored the `scanner` to loop through a list of checks. Alias ignore for check is only added to the list if `Alias` is set for a check.
- Will emit alias to auto generated documentation if set
- Alias added as an option in `tfsec-skeleton`
- Updated `CONTRIBUTING.md` to mention Alias in examples

 ### Done:
- [x] alias support
- [x] documentation update
- [x] make check generator update
- [x] unit tests
- [x] backward compatibility for checks without alias

### TODO (when solution roughly accepted)
- [ ] add in more aliases
- [ ] more benchmarks + maybe some optimizations

### Benchmarks
for 6144 blocks the diff in running checks is
- no alias 8.26s 
- alias 8.76s

for 12288 blocks the diff in running checks is
- no alias 38s 
- alias 40s

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,7 @@ Key attributes requested;
 
 - Provider: Select the provider from the list
 - Short Code: This is a very terse description of the check, it will form the check name
+- Alias: This is a alias that can be used in `tfsec:ignore` (aws-security-group-description-missing)
 - Summary: A slightly longer free text summary of the check
 - Required Types: What kind of blocks is this check for (resource, data, variable etc). Provide this as a space separated list
 - Required Label: What kind of labels is this check for (aws_instance, google_container_cluster). Provide this as a space separated list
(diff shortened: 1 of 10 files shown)


owenrumney:
Hey @michal-franc  - I'm not sure about this one tbh. I think it opens us up to inconsistency in the naming of the aliases at scale. 

Also, either way there is a need to go go back to the documentation to find out what the check is. As it stands, if the user takes the check code for the ignore they can add in a description as part of the comment if they wish. If the see the alias but don't know what check it maps to, the need to go looking to find out the specific code.

Will discuss with @liamg

owenrumney:
@michal-franc - we aren't going to go in this direction, but appreciate the time taken. Longer-term we will likely move away from codes to a consistent ID name which is more clear an moves us away from incremental IDs which present their own set of problems

PR #682, opened 2021-04-07
new flag include ignored and counting ignored checks

### LINKED ISSUE #664: Display ignored checks
When I ignore certain checks I would like to run command that would show me all the ignored checks.
New flag to collect ignored checks and display them.

michal-franc:
closes  #664

#### count ignored checks
When using `tfsec:ignore` we might forget about these ignores and it would be good to check how many of these are there to make a decision for instance that we have too many ignores in the project.
image

#### new flag `--include-ignored`
When checking where are the ignores we could actually try forcing display of them by `tfsec` thus `--include-ignored` flag might be helpful.

#### log ignored checks in debug log
When checking where are the ignores we could also check debug logs to check which rules are actually ignored.
image

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -39,6 +39,7 @@ var conciseOutput = false
 var excludeDownloaded = false
 var detailedExitCode = false
 var includePassed = false
+var includeIgnored = false
 var allDirs = false
 
 func init() {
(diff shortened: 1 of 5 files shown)


owenrumney:
Nice, like this

michal-franc:
Not sure @owenrumney i merging check the line with line above works correctly - could you add a test condition that would cover that scenario?

I added in the ignore above the block.

owenrumney:
@michal-franc  - added to this PR. Let me know you're happy and I'll merge

owenrumney (approved):

owenrumney (approved):

owenrumney:
@michal-franc  - I don't understand why I couldn't add my change to your branch. I've branched your branch in https://github.com/tfsec/tfsec/pull/683

PR #716, opened 2021-05-03
document using "provider" as a type

### LINKED ISSUE #711: Create checks for provider configuration
We want to enforce certain settings on provider configurations. More specifically, soon the [default tags](https://github.com/hashicorp/terraform-provider-aws/pull/19084) configuration on the provider level will become generally available and we'd like to enforce the use of that configuration. For example, we want to enforce that the `Application` tag is set:

```
provider "aws" {
  default_tags {
    tags = {
      Application = "Test"
    }
  }
}
```

It would be great if we could write custom checks that validate provider configuration. Looking at the [documentation](https://tfsec.dev/docs/custom_checks/) currently we can only check the `resource, data, module, variable` types.

For the specific use case mentioned above we found [this example](https://tfsec.dev/docs/check_library/tagged/) for checking tags which also mentions that you can use `aws_*` for all resource types. That doesn't work however as that will also enforce tags on resources that don't support it. Also more in general it would be great if this library can validate provider configurations as well.

SanderKnape:
Fixes #711 , apparently this was already possible

### DIFF
--- a/docs-website/docs/config/custom_checks.md
+++ b/docs-website/docs/config/custom_checks.md
@@ -81,7 +81,7 @@ The check contains up of the following attributes;
 |:-----------|:-------------|
 | code | The custom code that your check will be known as |
 | description | A description for the code that will be included in the output|
-|requiredTypes | The block types to apply the check to - resource, data, module, variable |
+|requiredTypes | The block types to apply the check to - provider, resource, data, module, variable |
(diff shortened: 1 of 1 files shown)


owenrumney (approved):

chewrocca:
I don't think this is displaying on the docs

https://aquasecurity.github.io/tfsec/v1.26.3/guides/configuration/custom-checks/#what-does-a-check-file-look-like 

I don't see this under requiredTypes.

PR #717, opened 2021-05-04
Added AZU028 Ensure the Function App can only be accessed via HTTPS. The default is false.

jpmicrosoft:
Added AZU028 Ensure the Function App can only be accessed via HTTPS. The default is false.


owenrumney:
@jpmicrosoft its failing the documentation check - we've added `Impact` and `Resolution` - see https://github.com/tfsec/tfsec/blob/master/internal/app/tfsec/checks/aws001.go#L38-L39 for an example

jpmicrosoft:
@owenrumney I added added Impact and Resolution. Thanks.

owenrumney:
sorry, I wasn't clear, you need to define them too - see https://github.com/tfsec/tfsec/blob/master/internal/app/tfsec/checks/aws001.go#L12-L13 for an example :-)

owenrumney (approved):

PR #735, opened 2021-05-16
adding the first Oracle Cloud check 

louwersj:
Adding the first Oracle Cloud check, it will check on the use of an IP from the external IP pool versus an IP from the internal IP pool. Warning for the use of publicly routable IPs which could expose a customer service to the public internet.

### DIFF
--- a/internal/app/tfsec/checks/oci001.go
+++ b/internal/app/tfsec/checks/oci001.go
@@ -1,0 +1,66 @@
+package checks
+
+import (
+	"fmt"
+
+	"github.com/tfsec/tfsec/internal/app/tfsec/parser"
+	"github.com/tfsec/tfsec/internal/app/tfsec/scanner"
+)
+
+const OCIComputeIpReservation scanner.RuleCode = "OCI001"
+const OCIComputeIpReservationDescription scanner.RuleSummary = "Compute instance requests an IP reservation from a public pool"
+const OCIComputeIpReservationImpact = "The compute instance has the ability to be reached from outside"
+const OCIComputeIpReservationResolution = "Reconsider the use of an public IP"
+const OCIComputeIpReservationExplanation = `
+Compute instance requests an IP reservation from a public pool
+
+The compute instance has the ability to be reached from outside, you might want to sonder the use of a non public IP.
+`
+const OCIComputeIpReservationBadExample = `
+resource "opc_compute_ip_address_reservation" "my-ip-address" {
+	name            = "my-ip-address"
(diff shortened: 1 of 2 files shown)


internal/app/tfsec/checks/oci001.go
```diff
@@ -0,0 +1,66 @@
...
+				"https://registry.terraform.io/providers/hashicorp/opc/latest/docs/resources/opc_compute_ip_address_reservation",
+				"https://registry.terraform.io/providers/hashicorp/opc/latest/docs/resources/opc_compute_instance",
+			},
+		},
+		Provider:       scanner.OracleProvider,
+		RequiredTypes:  []string{"resource"},
+		RequiredLabels: []string{"opc_compute_ip_address_reservation"},
+		CheckFunc: func(check *scanner.Check, block *parser.Block, _ *scanner.Context) []scanner.Result {
+			if attr := block.GetAttribute("ip_address_pool"); attr != nil {
+				if attr.IsAny("public-ippool") {
```
owenrumney, line 52: If there is only one value with is an issue you can use `attr.Equals("public-ippool", parser.IngnoreCase)`

owenrumney (commented):
Looks good thanks, could you add tests like the other checks too?

internal/app/tfsec/checks/oci001.go
louwersj: Found the testing pieces... I will start writing one... 

Please do have some patience with me, it is a learning curve which we need to go through. As soon as we have nailed this I think we can get some people within my team who can leverage the learning curve I am taking right ow and help adding new checks ;)

owenrumney: Hey @louwersj no pressure at all, really appreciate you helping out

louwersj: Added the first attempt to write a test for the OCI001 check...

louwersj (commented):
Test added.

internal/app/tfsec/checks/oci001.go
owenrumney: Hey @louwersj  - looking at the docs, it seems like `ip_address_pool` attribute is better as an `Equals`

louwersj:
@owenrumney can you please have a look at the inclusion of the test? I think with adding the test the first OCI check (and test) should be OK for a merge. Currently the merge is awaiting your reviewers approval.

owenrumney (approved):
looks great, thanks

PR #736, opened 2021-05-17
[skip ci] Use official Docker image in docker run

atombrella:
Seems a bit weird why there's a list of official images, but then tell people to use people to use an "unofficial" image (I know a maintainer made it, but the tfsec namespace is better) that hasn't been updated in the last 3 months :)


owenrumney (approved):
I've meant to do this countless times, thanks!!

PR #767, opened 2021-06-02
Refactoring and several minor bugfixes

liamg:
- Added `--stop-on-hcl-error` flag to cause tfsec to stop and exit with a non-zero status code if it encounters invalid HCL. The HCL parse error will be output to stdout.
- Fixed multiple test bugs which caused invalid HCL (e.g. missing brace) in test and example code to cause tests to incorrectly pass.
- Removed "special" types for IDs, descriptions etc.
- Renamed "checks" to "rules" where appropriate.
- Results can now contain more than one informational link (where supported by output format).
- Check functions now have a simpler signature, and returning single/multiple results is no longer as clumsy.
- Functional options are now in use by the scanner and parser.
- External scanner can now return passing results.
- Reorganised internal/external packages.


cmd/tfsec/main.go
```diff
@@ -47,14 +51,16 @@ var ignoreWarnings = false
 var ignoreInfo = false
 var allDirs = false
 var runStatistics bool
+var stopOnHCLError bool
```
owenrumney-f3, line 54: I think this ought to be default to true and changed to `ignore-hcl-errors`? You'd need to know you had an hcl to stop on them

liamg: I think so too, was worried about breaking BC though - people's builds will now start failing if they have invalid HCL in there. But maybe that is a good thing... :sweat_smile:

owenrumney-f3: there are a few scenarios -

1. user has broken terraform which has hcl error and they don't know (not sure how because it probably wouldn't deploy)
2. tfsec stops on error incorrectly, we get an issue and look at it on merit

owenrumney (approved):

PR #768, opened 2021-06-07
Add a check for IAM policies for KMS

felipesere:
AWS SecurityHub flags policies that grant permissions
to all possible KMS keys. This falls under the umbrella of "least
privilege" as having such broad permissions would allow an entity
to decrypt data they are not privy to.

### DIFF
--- a/internal/app/tfsec/rules/aws097.go
+++ b/internal/app/tfsec/rules/aws097.go
@@ -1,0 +1,96 @@
+package rules
+
+import (
+	"fmt"
+	"github.com/tfsec/tfsec/pkg/severity"
+
+	"github.com/tfsec/tfsec/pkg/result"
+
+	"github.com/tfsec/tfsec/pkg/provider"
+
+	"github.com/tfsec/tfsec/internal/app/tfsec/hclcontext"
+
+	"github.com/tfsec/tfsec/internal/app/tfsec/block"
+
+	"github.com/tfsec/tfsec/pkg/rule"
+
+	"github.com/tfsec/tfsec/internal/app/tfsec/scanner"
+)
+
+const AWSKMSManagedPoliciesShouldNotAllowDecryptionActionsOnAllKMSKeys = "AWS097"
+const AWSKMSManagedPoliciesShouldNotAllowDecryptionActionsOnAllKMSKeysDescription = "IAM customer managed policies should not allow decryption actions on all KMS keys"
+const AWSKMSManagedPoliciesShouldNotAllowDecryptionActionsOnAllKMSKeysImpact = "Identities may be able to decrypt data which they should not have access to"
+const AWSKMSManagedPoliciesShouldNotAllowDecryptionActionsOnAllKMSKeysResolution = "Scope down the resources of the IAM policy to specific keys"
+const AWSKMSManagedPoliciesShouldNotAllowDecryptionActionsOnAllKMSKeysExplanation = `
+IAM policies define which actions an identity (user, group, or role) can perform on which resources. Following security best practices, AWS recommends that you allow least privilege. In other words, you should grant to identities only the kms:Decrypt or kms:ReEncryptFrom permissions and only for the keys that are required to perform a task.
+`
+const AWSKMSManagedPoliciesShouldNotAllowDecryptionActionsOnAllKMSKeysBadExample = `
+data "aws_iam_policy_document" "kms_policy" {
+  statement {
+    principals {
+      type        = "AWS"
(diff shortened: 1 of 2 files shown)


felipesere:
This is the first time I'm contributing a check like this so I'd be keen to see if I missed anything.
Also, I'm worried the pattern might be too broad/not broad enough?

internal/app/tfsec/rules/aws046.go
```diff
@@ -71,19 +67,18 @@ func init() {
 
 			if statementBlocks := block.GetBlocks("statement"); statementBlocks != nil {
 				for _, statementBlock := range statementBlocks {
-					if effect := statementBlock.GetAttribute("effect"); effect != nil {
-						if effect.Type() == cty.String && strings.ToLower(effect.Value().AsString()) == "deny" {
-							continue
-						}
+					if hasEffectDeny(statementBlock) {
```
owenrumney, line 70: I appreciate it goes against the principle of DRY, but we want to keep the checks atomic so they can be deprecated with ease so trying to avoid functions spanning multiple checks.

As it is, this can be achieved with 

```
if statementBlock.HasChild("effect") && statementBlock.GetAttribute("effect").Equals("deny", block.IgnoreCase) {
	continue
}
```

internal/app/tfsec/rules/aws097.go
```diff
@@ -0,0 +1,120 @@
...
+		Provider:       provider.AWSProvider,
+		RequiredTypes:  []string{"data"},
+		RequiredLabels: []string{"aws_iam_policy_document"},
+		CheckFunc: func(set result.Set, block *block.Block, _ *hclcontext.Context) {
+
+			if statementBlocks := block.GetBlocks("statement"); statementBlocks != nil {
+				for _, statementBlock := range statementBlocks {
+
+					// Denying a broad set of KMS access is fine
+					if hasEffectDeny(statementBlock) {
```
owenrumney, line 79: can be simplified to 
```
if statementBlock.HasChild("effect") && statementBlock.GetAttribute("effect").Equals("deny", block.IgnoreCase) {
	continue
}
```

internal/app/tfsec/rules/aws097.go
```diff
@@ -0,0 +1,120 @@
...
+						continue
+					}
+
+					if containsActionLike(statementBlock, "kms") {
```
owenrumney, line 83: This can be simplified to then w e don't need either of the additional functions

```
if statementBlock.HasChild("actions") || statementBlock.GetAttribute("actions").Contains("kms") {
	if resources := statementBlock.GetAttribute("resources"); resources != nil {
		if resources.Contains("*") {
			set.Add(
				result.New().
					WithDescription(fmt.Sprintf("Resource '%s' a policy with KMS actions for all KMS keys.", b.FullName())).
					WithRange(b.Range()).
					WithAttributeAnnotation(resources).
					WithSeverity(severity.Error),
			)
		}
	}
}
```

owenrumney (commented):
A couple of minor points - we want to use the helper methods as much as possible so we can move to an improved way of defining the checks. 

WRT to the pattern, it looks okay to me on scope, and I agree its an error rather than warning

owenrumney:
I appreciate my use of the work "simplified" might be open for argument - I mean more that it is in keeping with self-contained checks

internal/app/tfsec/rules/aws046.go
felipesere: Oh I can totally get behind this! 👍

internal/app/tfsec/rules/aws097.go
felipesere: 👍

felipesere:
I took out the changes to `aws046.go` to keep this PR focused.

On a tangential note: We work a lot with SecurityHub and I wonder if there is value in somehow tagging rules with "AWS SecurityHub" and bringing over more of their controls?

owenrumney (approved):

owenrumney:
Thanks for this @felipesere - it'll take a short while to make it out as a release while we verify some refactoring

With regard to Security hub, definitely bringing some more of the controls over would be ideal - as for tagging rules, we don't have a labelling concept as such, will give that some thought

felipesere:
Just ran this locally against our wider repo and got some errors. I'll be debugging this and putting up a fix when I figure it out.
Its something with `resources.Contains("*")` not working for references like
```
resource = [
  var.some_arn
]
```

owenrumney:
Hmm, I thought contains was evaluating that, will dig into that a bit from my side too

felipesere:
Found the broad "issue": https://github.com/tfsec/tfsec/pull/770

PR #770, opened 2021-06-07
Dynamic types don't really have a value that can 'contain' a substring

felipesere:


owenrumney:
@felipesere  - can I confirm that we're talking about variables that have no default an a tfvars has not been supplied? tfsec will attempt to resolve the variable if the info is there, otherwise yes, it should be false when not available

felipesere:
Yeah exactly. Should add some tests around block/attribute to make the distinction and intention more clear? 😄

owenrumney (approved):

PR #782, opened 2021-06-11
The check for ignoring should use the resource block

### LINKED ISSUE #781: tfsec `v0.40.01` ignore rules inconsistent - raises error code on `AWS093` and `AWS032`
**Describe the bug**
Our CI builds started failing yesterday with the release of tfsec v0.40.0. Where v0.39.42 ran fine previously, the recently released v0.40.0 and v0.40.1 are causing us issues. Specifically, we were ignoring `AWS093` and `AWS032` but started seeing their respective error messages pop up when running tfsec in our CI builds, which leads to failures. I installed the latest version via brew (`v0.40.1`) and verified locally the same error is occurring. 

Further testing indicates that the rules are successfully ignored when passed through the command line using the `-e` flag but if they are added into the terraform file via `#tfsec:ignore:AWS093`, they aren't being ignored.

**To Reproduce**
1. Install tfsec `v0.40.01` or up
2. Run `tfsec` against a terraform file which contains `#tfsec:ignore:AWS093` -> see an error
3. Try running `tfsec -e AWS093` against that same file -> see no error as expected

**Expected behavior**
tfsec should ignore the rules as defined using the `#tfsec:ignore:XXXXXX` format

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux docker images, Mac OS

**Additional context**

Below is a snippet of the output with our code removed:

```sh
Check 1

  [AWS093][INFO] Resource 'aws_ecr_repository.bert' does not have CMK encryption configured
  /Users/johannes.giorgis/src/professional/ernie/services/bert/main.tf:27-35


<...relevant code snippet skipped...>

  Impact:     Using AWS managed keys does not allow for fine grained control
  Resolution: Use customer managed keys


Check 2

  [AWS093][INFO] Resource 'aws_ecr_repository.bert' should have the encryption type set to KMS
  :0
```

owenrumney:
- take the range details from the resource block rather than the result
block

- add a check for someone adding the tfsec ignore in about the result
block

Resolves #781

### DIFF
--- a/internal/app/tfsec/scanner/scanner.go
+++ b/internal/app/tfsec/scanner/scanner.go
@@ -85,7 +85,7 @@ func (scanner *Scanner) Scan(blocks []*block.Block) []result.Result {
 }
 
 func (scanner *Scanner) checkRangeIgnored(id string, r block.Range, b block.Range) bool {
-	raw, err := ioutil.ReadFile(r.Filename)
+	raw, err := ioutil.ReadFile(b.Filename)
 	if err != nil {
 		return false
 	}
@@ -110,7 +110,7 @@ func (scanner *Scanner) checkRangeIgnored(id string, r block.Range, b block.Rang
 		}
 	}
 
-	// check the line above the block
+	// check the line above the actual resource block
 	if b.StartLine-1 > 0 {
 		line := lines[b.StartLine-1]
 		if ignored := checkLineForIgnore(line, ignoreAll, ignoreCode); ignored {


internal/app/tfsec/scanner/scanner.go
```diff
@@ -111,6 +111,14 @@ func (scanner *Scanner) checkRangeIgnored(id string, r block.Range, b block.Rang
 	}
 
 	// check the line above the block
+	if r.StartLine-1 > 0 {
```
liamg, line 114: It looks like this block already exists below 🤔

liamg: Nope, I can't read

liamg (dismissed):

owenrumney:
Yup @liamg  - that seems to have been unnecessary after all. The key part is the range filename

liamg (approved):
Ah I see!

PR #798, opened 2021-06-23
Fix --include-passed

knqyf263:
`--include-passed` doesn't work now.

```
$ tfsec --include-passed /path/to/tf
```

### DIFF
--- a/internal/app/tfsec/scanner/scanner.go
+++ b/internal/app/tfsec/scanner/scanner.go
@@ -63,8 +63,9 @@ func (scanner *Scanner) Scan(blocks []*block.Block) []result.Result {
 				if rule.IsRuleRequiredForBlock(r, checkBlock) {
 					debug.Log("Running rule for %s on %s.%s (%s)...", r.ID, checkBlock.Type(), checkBlock.FullName(), checkBlock.Range().Filename)
 					ruleResults := rule.CheckRule(r, checkBlock, context)
-					if scanner.includePassed && ruleResults == nil {
-						res := result.New(checkBlock).WithRange(checkBlock.Range()).WithStatus(result.Passed).WithSeverity(severity.None)
+					if scanner.includePassed && ruleResults.All() == nil {
+						res := result.New(checkBlock).WithRuleID(r.ID).WithDescription(r.Documentation.Summary).
+							WithRange(checkBlock.Range()).WithStatus(result.Passed).WithSeverity(severity.None)
 						results = append(results, *res)
 					} else if ruleResults != nil {
 						for _, ruleResult := range ruleResults.All() {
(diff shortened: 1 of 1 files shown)


internal/app/tfsec/scanner/scanner.go
```diff
@@ -63,8 +63,9 @@ func (scanner *Scanner) Scan(blocks []*block.Block) []result.Result {
 				if rule.IsRuleRequiredForBlock(r, checkBlock) {
 					debug.Log("Running rule for %s on %s.%s (%s)...", r.ID, checkBlock.Type(), checkBlock.FullName(), checkBlock.Range().Filename)
 					ruleResults := rule.CheckRule(r, checkBlock, context)
-					if scanner.includePassed && ruleResults == nil {
-						res := result.New(checkBlock).WithRange(checkBlock.Range()).WithStatus(result.Passed).WithSeverity(severity.None)
+					if scanner.includePassed && ruleResults.All() == nil {
```
owenrumney, line 66: Agree with this, thanks

internal/app/tfsec/scanner/scanner.go
```diff
@@ -63,8 +63,9 @@ func (scanner *Scanner) Scan(blocks []*block.Block) []result.Result {
...
+						res := result.New(checkBlock).WithRuleID(r.ID).WithDescription(r.Documentation.Summary).
```
owenrumney, line 67: This addition of `WithDescription` isn't correct - the Description is specific to the resource, eg; `Resource 'aws_instance.bastion` has a public IP` sort of thing. Documentation.Summary is the overall rule summary `AWS Instances shouldn't have Public IPs` that sort of thing. 

These Passed checks are odd in that they've not failed anything, but if anything the Description should be along the lines of `WithDescription(fmt.Sprintf("Resource '%s' had no issues identified", checkBlock.FullName()))`

liamg: I agree, I think the `no issues identified` idea is pretty clean:
```suggestion
						res := result.New(checkBlock).WithRuleID(r.ID).WithDescription(fmt.Sprintf("Resource '%s' had no issues identified", checkBlock.FullName())).
```

knqyf263: OK, tfsec used to display the rule summary like "[AWS004][PASSED] Use of plain HTTP.", and this bug is a regression. I just complied with the past behavior.

image

`Result.Description` is shown in a scan result, so I inserted Documentation.Summary into Description.
https://github.com/tfsec/tfsec/blob/988cedd3f1f06f703f09ac4da56834965c83857c/internal/app/tfsec/formatters/default.go#L76

But can we still keep the rule summary somewhere in the Result struct like WithRuleSummary? That is because "[AWS003] Resource aws_alb_listener had no issues identified" doesn't have enough information for users. I think a user wants to know what kind of checks passed.

```suggestion
						res := result.New(checkBlock).WithRuleID(r.ID).WithDescription(fmt.Sprintf("Resource '%s' had no issues identified", checkBlock.FullName()).WithRuleSummary(r.Documentation.Summary).
```

owenrumney:
```suggestion
						res := result.New(checkBlock).WithRuleID(r.ID).WithDescription(fmt.Sprintf("Resource '%s' passed check: %s", checkBlock.FullName(), r.Documentation.Summary))
```

The summary would need to be in the description in this case, so I think this suggestion would do as it yields

image

knqyf263: fixed!

owenrumney (approved):

PR #803, opened 2021-06-28
traverse the load balancer to check type

owenrumney:
- add attribute functions to walk the scope and find the related
  resource to a reference so further checks can be performed
- rename `ReferencesDataBlock` to `IsDataBlockReference` to make the intent clearer


liamg (approved):

liamg (approved):
LGTM

PR #812, opened 2021-06-29
Use the latest TLS policy in AWS021

sergei-ivanov:
AWS have just updated their CloudFront TLS policy to drop non-GCM cipher suites. The latest policy is now `TLSv1.2_2021`.
https://aws.amazon.com/about-aws/whats-new/2021/06/amazon-cloudfront-announces-new-tlsv12_2021-security-policy-for-viewer-connections/


owenrumney (approved):
Thanks @sergei-ivanov

PR #813, opened 2021-07-01
Keep the check for rule AWS021 consistent with the error message

bitsapien:
The error message for AWS021 says -
"You should not use outdated/insecure TLS versions for
encryption. You should be using TLS v1.2+."

The rule is now consistent with this version and will not
complain if you afix a version higher than what tfsec prescribes.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/secure-connections-supported-viewer-protocols-ciphers.html#secure-connections-supported-ciphers


liamg:
I don't think we'll need this as the next iteration is likely to be a TLS 1.3 policy rather than 1.2 2022 etc.

PR #822, opened 2021-07-06
Fix some cyclo issues

owenrumney:
- add GetAttribute to do nested attributes

### DIFF
--- a/internal/app/tfsec/block/hclblock.go
+++ b/internal/app/tfsec/block/hclblock.go
@@ -197,9 +197,6 @@ func (block *HCLBlock) GetAttributes() []Attribute {
 }
 
 func (block *HCLBlock) GetAttribute(name string) Attribute {
-	if strings.Contains(name, "/") {
-		return block.GetNestedAttribute(name)
-	}
 	if block == nil || block.hclBlock == nil {
 		return nil
 	}
@@ -216,12 +213,12 @@ func (block *HCLBlock) GetNestedAttribute(name string) Attribute {
 	blocks := parts[:len(parts)-1]
 	attrName := parts[len(parts)-1]
 
-	working := block
+	var working Block = block
 	for _, b := range blocks {
 		if checkBlock := working.GetBlock(b); checkBlock == nil {
 			return nil
 		} else {
-			working = checkBlock.(*HCLBlock)
+			working = checkBlock
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


internal/app/tfsec/rules/aws004.go
```diff
@@ -60,49 +58,57 @@ func init() {
...
+
+			protocolAttr := resourceBlock.GetAttribute("protocol")
+
+			if protocolAttr != nil {
+				if protocolAttr.IsResolvable() && protocolAttr.Equals("HTTPS") {
+					return
+				}
+				if protocolAttr.IsResolvable() && protocolAttr.Equals("HTTP") {
+					// check if this is a redirect to HTTPS - if it is, then no problem
+					if redirectProtocolAttr := resourceBlock.GetAttribute("default_action/redirect/protocol"); redirectProtocolAttr != nil {
```
owenrumney, line 74: @liamg  - Interested to know your thoughts on this

internal/app/tfsec/block/hclblock.go
```diff
@@ -208,6 +211,27 @@ func (block *HCLBlock) GetAttribute(name string) Attribute {
 	return nil
 }
 
+func (block *HCLBlock) GetNestedAttribute(name string) Attribute {
+	parts := strings.Split(name, "/")
+	blocks := parts[:len(parts)-1]
+	attrName := parts[len(parts)-1]
+
+	working := block
```
liamg, line 219: If we cast it to the interface we can remove the cast on line 224 and it'll be future-proofed.
```suggestion
	working := block.(Block)
```

internal/app/tfsec/block/hclblock.go
```diff
@@ -197,6 +197,9 @@ func (block *HCLBlock) GetAttributes() []Attribute {
 }
 
 func (block *HCLBlock) GetAttribute(name string) Attribute {
+	if strings.Contains(name, "/") {
```
liamg, line 200: Can block names definitely not contain `/`? If they can we could just ask that people call `GetNestedAttribute` directly.

internal/app/tfsec/rules/aws004.go
liamg: Looks good, I like it :+1:

liamg (changes requested):

internal/app/tfsec/block/hclblock.go
owenrumney: I don't think so, but I agree it should be an explicit action so have fixed

internal/app/tfsec/block/hclblock.go
owenrumney: Done

liamg (approved):

PR #832, opened 2021-07-08
Allow KMS actions for all resources for non-IAM policy documents

### LINKED ISSUE #794: AWS097 kms key policy should only apply to IAM policies not KMS key policies
This may be a bug that's unable to be worked due to filtering in terraform, documentation may be able to be updated to reference IAM policies instead of key policies.

tfsec check AWS097 applies to all aws_iam_policy_document resources but AWS documentation has the resource of "*" required. This should be only for policies attached to IAM resources rather than KMS keys.

AWS documentation around KMS keys requiring resource of "*"

Documentation on https://tfsec.dev/docs/aws/AWS097/ requiring KMS key policies to not be on resource "*" for AWS IAM policies

https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#fsbp-kms-1

Terraform has a circular dependency map with the suggested fix from https://tfsec.dev/docs/aws/AWS097/

Steps to reproduce the behaviour:
Run a terraform plan across the following code to produce a cycle dependency, a KMS key policy can't be attached to a KMS key policy until the KMS key is created, which requires the kms key policy:

```
resource "aws_kms_key" "key" {
  description         = "tfsec example"
  policy              = data.aws_iam_policy_document.policy.json
}

data "aws_iam_policy_document" "policy" {
  statement {
    sid = "Admins"

    effect = "Allow"

    actions = ["kms:*"]

    resources = [aws_kms_key.key.arn]

    principals {
      type        = "AWS"
      identifiers = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"
    }
  }
}
```
Error:
```
Error: Cycle: data.aws_iam_policy_document.policy, aws_kms_key.key
```

KMS actions should apply only to IAM related policies rather than KMS related resources from https://docs.aws.amazon.com/kms/latest/developerguide/control-access-overview.html#overview-policy-elements

liamg:
Resolves #794

### DIFF
--- a/internal/app/tfsec/block/attribute.go
+++ b/internal/app/tfsec/block/attribute.go
@@ -30,4 +30,7 @@ type Attribute interface {
 	IsResourceBlockReference(resourceType string) bool
 	ReferencesBlock(b Block) bool
 	IsResolvable() bool
+	IsString() bool
+	IsNumber() bool
+	IsBool() bool
 }
--- a/internal/app/tfsec/block/hclattribute.go
+++ b/internal/app/tfsec/block/hclattribute.go
@@ -36,6 +36,18 @@ func (attr *HCLAttribute) Type() cty.Type {
 	return attr.Value().Type()
 }
 
+func (attr *HCLAttribute) IsString() bool {
+	return attr.Value().Type() == cty.String
+}
+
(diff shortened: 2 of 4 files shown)


owenrumney (approved):
NIce

PR #846, opened 2021-07-09
Rework GCP-001 (and add new raw_key check)

### LINKED ISSUE #137: GCP001 & GCP002 Checks are not applicable
## GCP001 & GCP002 Checks are not applicable
 I've looked at these two rules and believe they may not be relevant checks in their current form. Based on the documentation and some testing, google automatically encrypts disks and buckets if no client-managed encryption is specified.

In some general documentation it is stated [buckets](https://cloud.google.com/storage/docs/encryption) and [disks](https://cloud.google.com/compute/docs/disks/customer-supplied-encryption) are encrypted by default. The options both checks are looking for are used to define customer managed encryption keys.

For disks if you do create the `disk_encryption_key` block and leave it blank, it will [automatically generate keys](https://github.com/terraform-providers/terraform-provider-google/blob/5dfc9b8b72090b9a7c36cb80ad0394b2db43e108/google/resource_compute_disk.go#L274). Buckets are similar in that if you do not specify any encryption it defaults to google managed keys.

### Possible actions
Not sure if deleting a rule is the best course of action. One possibility could be modifying these to warn in the event someone is specifying customer managed keys. ie: remind them to ensure they are securely managing their keys, or to check that the specified `default_kms_key_name` is accessible where the object is being created.

liamg:
Resolves #137

### DIFF
--- a/internal/app/tfsec/rules/gcp013.go
+++ b/internal/app/tfsec/rules/gcp013.go
@@ -1,0 +1,79 @@
+package rules
+
+import (
+	"fmt"
+
+	"github.com/tfsec/tfsec/internal/app/tfsec/block"
+	"github.com/tfsec/tfsec/internal/app/tfsec/hclcontext"
+	"github.com/tfsec/tfsec/internal/app/tfsec/scanner"
+	"github.com/tfsec/tfsec/pkg/provider"
+	"github.com/tfsec/tfsec/pkg/result"
+	"github.com/tfsec/tfsec/pkg/rule"
+	"github.com/tfsec/tfsec/pkg/severity"
+)
+
+const GCPRawEncryptionKeySpecifiedForComputeDisk = "GCP013"
+const GCPRawEncryptionKeySpecifiedForComputeDiskDescription = "The encryption key used to encrypt a compute disk has been specified in plaintext."
+const GCPRawEncryptionKeySpecifiedForComputeDiskImpact = "The encryption key should be considered compromised as it is not stored securely."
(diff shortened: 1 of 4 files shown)


internal/app/tfsec/rules/gcp013.go
```diff
@@ -0,0 +1,79 @@
...
+		ID: GCPRawEncryptionKeySpecifiedForComputeDisk,
+		Documentation: rule.RuleDocumentation{
+			Summary:     GCPRawEncryptionKeySpecifiedForComputeDiskDescription,
+			Explanation: GCPRawEncryptionKeySpecifiedForComputeDiskExplanation,
+			Impact:      GCPRawEncryptionKeySpecifiedForComputeDiskImpact,
+			Resolution:  GCPRawEncryptionKeySpecifiedForComputeDiskResolution,
+			BadExample:  GCPRawEncryptionKeySpecifiedForComputeDiskBadExample,
+			GoodExample: GCPRawEncryptionKeySpecifiedForComputeDiskGoodExample,
+			Links: []string{
+				"https://cloud.google.com/compute/docs/disks/customer-supplied-encryption",
```
owenrumney, line 48:
```suggestion
                                "https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk#kms_key_self_link",
				"https://cloud.google.com/compute/docs/disks/customer-supplied-encryption",
```

owenrumney (approved):

PR #862, opened 2021-07-13
Fix default formatter output for critical results

liamg:


internal/app/tfsec/formatters/default.go
```diff
@@ -19,10 +19,11 @@ import (
...
 
 var severityFormat = map[severity.Severity]string{
-	severity.Low:    tml.Sprintf("<white>%s</white>", severity.Low),
-	severity.Medium: tml.Sprintf("<yellow>%s</yellow>", severity.Medium),
-	severity.High:   tml.Sprintf("<red>%s</red>", severity.High),
-	"":              tml.Sprintf("<white>%s</white>", severity.Low),
+	severity.Low:      tml.Sprintf("<white>%s</white>", severity.Low),
+	severity.Medium:   tml.Sprintf("<yellow>%s</yellow>", severity.Medium),
+	severity.High:     tml.Sprintf("<red>%s</red>", severity.High),
+	severity.Critical: tml.Sprintf("<red>%s</red>", severity.Critical),
```
owenrumney, line 25: Is `BrightRed` an option?

liamg: There is `lightred`, so i'll flip them round :+1:

owenrumney: Nice

liamg: Used bold instead, worked better

owenrumney (approved):

PR #900, opened 2021-07-20
v1 skeleton update

owenrumney:
- Update the skeleton generator

### DIFF
--- a/cmd/tfsec-skeleton/templates.go
+++ b/cmd/tfsec-skeleton/templates.go
@@ -1,42 +1,38 @@
 package main
 
-const checkTemplate = `package rules
+const checkTemplate = `package {{ .Service}}
 
 import (
-	"github.com/aquasecurity/tfsec/internal/app/tfsec/block"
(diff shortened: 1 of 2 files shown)


cmd/tfsec-skeleton/templates.go
```diff
@@ -54,15 +50,16 @@ func init() {
...
 
 import (
 	"testing"
 
-	"github.com/aquasecurity/tfsec/internal/app/tfsec/rules"
+	"github.com/aquasecurity/tfsec/internal/app/tfsec/testutil"
 )
 
 func Test_{{.CheckName}}(t *testing.T) {
+	expectedCode := "{{ .ShortCode}}"
```
liamg, line 62: You want `provider-service-shortcode` here

liamg (changes requested):

cmd/tfsec-skeleton/templates.go
owenrumney: Fixed - `{{ .FullCode}}`

liamg (approved):

PR #909, opened 2021-07-20
Add support for the workspace ignores

### LINKED ISSUE #907: Ignore checks per workspace
There is no way to ignore a check for a specific workspace.

Same as the `--tfvars-file`, I'd like to be able to ignore a test in an environment but not on another maybe via flags or by taking into account the current workspace.

No alternatives have been considered.

**Additional context**
It's common to use workspaces to manage different environments. One may want to ignore a test (i.e.`GCP009`) in staging but not in production.

Right now, by adding the (i.e.) `#tfsec:ignore:GCP009` to the tf code it'll be applied to every workspace, creating a situation where you are forced to have a staging/production parity environment, which not everyone wants to.

owenrumney:
Pass the --workspace argument at runtime
ignores should include #tfsec:ignore:AWS006:ws:workspaceName

Resolves #907

### DIFF
--- a/internal/app/tfsec/scanner/scanner.go
+++ b/internal/app/tfsec/scanner/scanner.go
@@ -25,6 +25,7 @@ type Scanner struct {
 	includeIgnored    bool
 	excludedRuleIDs   []string
 	ignoreCheckErrors bool
+	workspaceName     string
 }
 
 // New creates a new Scanner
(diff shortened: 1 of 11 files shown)


internal/app/tfsec/scanner/scanner.go
```diff
@@ -153,12 +154,43 @@ func (scanner *Scanner) checkRangeIgnored(id string, r block.Range, b block.Bloc
...
-		return isIgnoreWithinExpiry(ignoreLine, id, b.Range())
+		return scanner.verifyIgnoreConditions(ignoreLine, id, b)
 	}
 
 	return foundValidIgnore
 }
 
+func (scanner *Scanner) verifyIgnoreConditions(ignoreLine string, id string, b block.Block) bool {
+	for _, ignoreCode := range strings.Split(ignoreLine, " ") {
+		if strings.Contains(ignoreCode, id) || strings.Contains(ignoreCode, "ignore:*") {
```
liamg, line 165: I think this is going to match sometimes when it shouldn't like `ignore:aws-123` will match on `Contains(ignore:aws-1)`

liamg (changes requested):

internal/app/tfsec/scanner/scanner.go
owenrumney: done

liamg (approved):

PR #918, opened 2021-07-21
Fix false positive IAM resource wildcard checks - resolves #915

### LINKED ISSUE #915: AWS099 False positive for actions without required resource specification
**Describe the bug**
AWS099 - IAM policy should avoid use of wildcards and instead apply the principle of least privilege

From https://docs.aws.amazon.com/service-authorization/latest/reference/list_identityandaccessmanagement.html:
> If there is no value for this column, you must specify all resources ("*") in the Resource element of your policy statement. 

**To Reproduce**
```terraform
resource "aws_iam_policy" "policy" {
  name        = "mfa-policy"

  policy = <<EOT
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": [
        "iam:ListVirtualMFADevices"
      ],
      "Effect": "Allow",
      "Resource": "*"
    }
  ]

}
EOT
}
```

**Additional context**
Fails with `guardduty:AcceptInvitation` action and `*` as resource too, did not test with other services.

liamg:
Resolves #915


owenrumney (approved):
yikes!

PR #930, opened 2021-07-22
Add check for security contact phone

owenrumney:
update generator to support hyphenated services

### DIFF
--- a/internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go
+++ b/internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go
@@ -1,0 +1,77 @@
+package securitycenter
+
+import (
+	"fmt"
+
+	"github.com/aquasecurity/tfsec/pkg/result"
+	"github.com/aquasecurity/tfsec/pkg/severity"
+
+	"github.com/aquasecurity/tfsec/pkg/provider"
+
+	"github.com/aquasecurity/tfsec/internal/app/tfsec/hclcontext"
+
+	"github.com/aquasecurity/tfsec/internal/app/tfsec/block"
+
+	"github.com/aquasecurity/tfsec/pkg/rule"
+
+	"github.com/aquasecurity/tfsec/internal/app/tfsec/scanner"
+)
+
+func init() {
+	scanner.RegisterCheckRule(rule.Rule{
+		Service:   "security-center",
+		ShortCode: "set-required-contact-details",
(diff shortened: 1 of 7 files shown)


internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go
```diff
@@ -0,0 +1,77 @@
...
+				)
+				return
+			}
+
+			phoneAttr := resourceBlock.GetAttribute("phone")
+			if phoneAttr == nil || phoneAttr.IsEmpty() {
+				set.Add(
+					result.New(resourceBlock).
+						WithDescription(fmt.Sprintf("Resource '%s' does not have a phone number set for the security contact", resourceBlock.FullName())).
+						WithAttributeAnnotation(phoneAttr).
```
liamg, line 71: `phoneAttr` might be `nil` here

internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go
```diff
@@ -0,0 +1,77 @@
...
+						WithRange(phoneAttr.Range()),
```
liamg, line 72: same as above

liamg (changes requested):

internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go
owenrumney: but it won't be - because it's checked for above in the MissingChild check

internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go
owenrumney: but it won't be - because it's checked for above in the MissingChild check

owenrumney: but it won't be - because it's checked for above in the MissingChild check

internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go
liamg: sorry, my bad - i just saw the `if phoneAttr == nil || ` above.

liamg (dismissed):

liamg (approved):

PR #944, opened 2021-07-23
Add rule for google sql backups

liamg:


internal/app/tfsec/rules/google/sql/enable_backup_rule.go
```diff
@@ -0,0 +1,91 @@
...
+				)
+				return
+			}
+
+			if backupBlock := settingsBlock.GetBlock("backup_configuration"); backupBlock == nil {
+				set.Add(
+					result.New(resourceBlock).
+						WithDescription(fmt.Sprintf("Resource '%s' does not have backups enabled.", resourceBlock.FullName())),
+				)
+			} else if enabledAttr := backupBlock.GetAttribute("enabled"); enabledAttr == nil {
```
owenrumney, line 78: no return above, so backupBlock could be nil

internal/app/tfsec/rules/google/sql/enable_backup_rule.go
```diff
@@ -0,0 +1,91 @@
...
+				set.Add(
+					result.New(resourceBlock).
+						WithDescription(fmt.Sprintf("Resource '%s' does not have backups enabled.", resourceBlock.FullName())),
+				)
+			} else if enabledAttr.IsFalse() {
```
owenrumney, line 83: as above.

owenrumney (changes requested):

internal/app/tfsec/rules/google/sql/enable_backup_rule.go
owenrumney: I like doing 
```
if resourceBlock.MissingChild("backup_configuration") {
				set.Add(
					result.New(resourceBlock).
						WithDescription(fmt.Sprintf("Resource '%s' does not have backups enabled.", resourceBlock.FullName())),
return
)
```

owenrumney (approved):

PR #950, opened 2021-07-26
Rule for GCP IAM assigning privileged role to service account

liamg:


internal/app/tfsec/block/hclattribute.go
```diff
@@ -474,20 +474,56 @@ func (attr *HCLAttribute) Reference() (*Reference, error) {
...
+		return refs
+	}
+	ref, err := attr.Reference()
+	if err != nil {
+		return nil
+	}
+	return append(refs, ref)
+}
+
+func (attr *HCLAttribute) ReferencesInTemplate() []*Reference {
```
owenrumney, line 512: does it need to be public and if yes, should it be on the interface?

owenrumney (commented):
question

internal/app/tfsec/block/hclattribute.go
liamg: good point, it's private now

owenrumney (approved):

PR #960, opened 2021-07-27
Create the codes json file during docs build

owenrumney:
this will be downloaded by the vscode plugin for ignore resolution

### DIFF
--- a/cmd/tfsec-docs/extension_codes.go
+++ b/cmd/tfsec-docs/extension_codes.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"encoding/json"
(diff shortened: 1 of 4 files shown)


cmd/tfsec-docs/extension_codes.go
```diff
@@ -0,0 +1,58 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+type checkBlock struct {
```
liamg, line 9: I don't think this is used (might be wrong, reviewing on phone)

liamg (approved):

PR #975, opened 2021-07-29
Add usage doc

owenrumney:
add usage permenantly to the navigation doc

### DIFF
--- a/cmd/tfsec-docs/navigation_docs.go
+++ b/cmd/tfsec-docs/navigation_docs.go
@@ -15,7 +15,8 @@ const navDocsTemplate = `---
 - title: Getting Started
   state: open
   docs:
-    - installation
+  	- installation
+	- usage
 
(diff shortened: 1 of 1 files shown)


cmd/tfsec-docs/navigation_docs.go
```diff
@@ -16,6 +16,7 @@ const navDocsTemplate = `---
   state: open
   docs:
     - installation
+	- usage
```
liamg, line 19: This indenting is weird, or the github app is weird, can't tell which.

liamg (dismissed):

cmd/tfsec-docs/navigation_docs.go
owenrumney: it was something funky in my ide masking the indent

liamg (approved):

PR #1015, opened 2021-08-11
Add requiredSources functionality to filter modules

aidan-canva:
# Intent
Add a `requiredSources` filter block to custom rules to make checking modules simple and targeted.

# Problem
Currently there is no way to easily target specific modules for custom rules, especially so if there is no common label pattern across instances of the module.  For modules that are stored on a local path, matching filtering with a complex `matchSpec` does not work as the `source` value is relative to the root module's position and is not a constant value.

# Proposed Solution
Addition of a top level property called `requiredSources` which operates similarly to the `requiredLabels` property and targets a modules `source` attribute.  For modules that reference a local path in their `source` attribute, the path is resolved to a path relative to the current working directory.  This enables users to write a custom rule that targets a module regardless of where in the directory tree it is referenced.

As reference, imagine the following directory structure:
```
tf
├── apps
│   ├── applicationA
│   │   ├── common
│   │   └── nested
│   └── applicationB
└── common
```

Root modules defined in `tf/apps/applicationA/nested` and `tf/apps/applicationB` would reference the top level `common` directory using a `source` value of `../../../common` and `../../common` respectively.  With this path resolution implementation, it would be resolved to a consistent value of `common`.  This also ensures the `common` directory under `apps/applicationA/` does not incorrectly match.

# Notes
- I had to break up the `IsRuleRequiredForBlock` function due to a `gocyclo` complexity threshold.  That refactor is in a separate commit.


owenrumney (approved):
Thanks for this! Will be interesting to see how much this gets used!

PR #1036, opened 2021-08-17
Add check to avoid mixing usage of security_group_rule and security_group with inline rules

sgavinio:
## Description

Hello, this adds an AWS check to ensure that usage of security groups with inline rules and security group rule resources are not mixed.

Mixing Terraform standalone `security_group_rule` resource and `security_group` resource with inline ingress/egress rules results in rules being overwritten during Terraform apply. 

This is intended behavior as per [terraform](https://github.com/hashicorp/terraform/issues/11011#issuecomment-283076580) but can result in security group rules being overwritten if one is not careful. Big banner warning from terraform [security group docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group) :

Screen Shot 2021-08-18 at 7 58 15 AM

## Solution

This adds the check in go where we have  access to the `GetReferencedBlock` function. 

I'm not able to express how I can refer to "co-related" resources with the custom check language. Please let me know  if this is achievable, may have just missed it. Thank you 🙏 

## References

* https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group
* https://github.com/hashicorp/terraform/issues/11011#issuecomment-283076580

### DIFF
--- a/internal/app/tfsec/rules/aws/vpc/disallow_mixed_sgr_rule.go
+++ b/internal/app/tfsec/rules/aws/vpc/disallow_mixed_sgr_rule.go
@@ -47,7 +47,7 @@ resource "aws_security_group" "good_example_sg" {
 }
 `},
 			Links: []string{
-				"https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group",
+				"https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group#resource-aws_security_group",
 				"https://github.com/hashicorp/terraform/issues/11011#issuecomment-283076580",
 			},
 		},
--- a/internal/app/tfsec/rules/aws/vpc/disallow_mixed_sgr_rule.go
+++ b/internal/app/tfsec/rules/aws/vpc/disallow_mixed_sgr_rule.go
@@ -1,0 +1,74 @@
+package vpc
+
+import (
+	"fmt"
+	"github.com/aquasecurity/tfsec/internal/app/tfsec/block"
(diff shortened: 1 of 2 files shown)


internal/app/tfsec/rules/aws/vpc/disallow_mixed_sgr_rule.go
```diff
@@ -0,0 +1,74 @@
...
+  	security_group_id = aws_security_group.good_example_sg.id
+	type = "ingress"
+	cidr_blocks = ["10.0.0.0/16", "172.31.0.0/16"]
+}
+
+resource "aws_security_group" "good_example_sg" {
+}
+`},
+			Links: []string{
+				"https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group",
```
owenrumney, line 50: Can we change this link and target the message specifically 
```suggestion
				"https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group#resource-aws_security_group",
```

owenrumney (dismissed):
This is great, really appreciate your contribution

internal/app/tfsec/rules/aws/vpc/disallow_mixed_sgr_rule.go
sgavinio: Ack. Linked changed. Thank you.

owenrumney (approved):

sgavinio:
Thanks for merging @owenrumney 🙏 . Is there a regular schedule for cutting a release? Just want to factor in when we get this from the upstream.

Issue #1037, opened 2021-08-19
--no-color not fully working without color


yi2020:
**Describe the bug**
--no-color started outputting styling escape codes.

**To Reproduce**
Run tfsec with --no-color and redirect to a file:
```
docker run --rm -v "$(pwd):/src" aquasec/tfsec /src --no-color > test.txt
```

Now, open the file with vi or a similar editor (don't cat it in terminal) and you'll see the markers, like these:
```
  [aws-lambda-enable-tracing][^[[0m^[[97mLOW^[[39m^[[0m] Resource 'aws_lambda_function.test_lambda' uses default value for tracing_config.mode
```

**Expected behavior**
The terminal styling markers shouldn't be in the output.

PR #1042, opened 2021-08-21
fix: no-color for severity

michal-franc:
fixes: https://github.com/aquasecurity/tfsec/issues/1037

#### Problem
severityFormat map was initialized before user flags were parsed by main tfsec app. These flags control `tml` parsing. Initialization of severityFormat uses `tml` so it was done with `default` values of tml - by default `no-color` is not set.

#### Fix
Moved severityFormat map generation to `runtime` context within `FormatDefault` function which will use `no-color` flag set by the user.

Added check to only populate this map if its nil.

Example:

```
  Result 1

  [aws-ecs-enable-container-insight][HIGH] Resource 'aws_ecs_cluster.bad_example' does not have containerInsights enabled
  /home/michal-franc/Work/go-src/github.com/michal-franc/tfsec/bin/linux/test.tf:1-3


       1 | resource "aws_ecs_cluster" "bad_example" {
       2 |   	name = "services-cluster"
       3 | }
       4 |
       5 |

  Legacy ID:  AWS090
  Impact:     Not all metrics and logs may be gathered for containers when Container Insights isn't enabled
  Resolution: Enable Container Insights
```

### DIFF
--- a/internal/app/tfsec/formatters/default.go
+++ b/internal/app/tfsec/formatters/default.go
@@ -19,15 +19,16 @@ import (
 	"github.com/liamg/tml"
 )
 
-var severityFormat = map[severity.Severity]string{
-	severity.Low:      tml.Sprintf("<white>%s</white>", severity.Low),
-	severity.Medium:   tml.Sprintf("<yellow>%s</yellow>", severity.Medium),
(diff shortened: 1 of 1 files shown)


owenrumney (approved):
Thanks @michal-franc  - We seem to have done this at the same time.

Will take yours

PR #1043, opened 2021-08-21
format severities at runtime

### LINKED ISSUE #1037: --no-color not fully working without color

owenrumney:
need to calculate the severity formatting at runtime because no color doesn't disable tml formatting till tfsec has started.

Calculate once at start of formatter then use the map as before

Resolves #1037

### DIFF
--- a/internal/app/tfsec/formatters/default.go
+++ b/internal/app/tfsec/formatters/default.go
@@ -19,12 +19,16 @@ import (
 	"github.com/liamg/tml"
 )
 
-var severityFormat = map[severity.Severity]string{
-	severity.Low:      tml.Sprintf("<white>%s</white>", severity.Low),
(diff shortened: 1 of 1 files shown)


owenrumney:
Closing, duplicates #1042

PR #1056, opened 2021-08-30
fix: do not ignore allExcludedRuleIDs params

### LINKED ISSUE #1055: --exclude options is ignored
**Describe the bug**
When scanning a project, the `--exclude` flags is not taken into account.

**To Reproduce**
1. Choose a rule, e.g. `aws-cloudwatch-log-group-customer-key`
2. Create a resource that does not respect this rule
3. run `tfsec . --exclude  aws-cloudwatch-log-group-customer-key` and see it fail.

**Expected behavior**
The scanning should not display errors / potential problems that are excluded.

**Additional context**
- it seems to me that it is directly related to https://github.com/aquasecurity/tfsec/pull/1053. I will create a fix PR.

p9f:
I believe this line has been forgotten in https://github.com/aquasecurity/tfsec/pull/1053

Resolves #1055

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -378,6 +378,8 @@ func getScannerOptions() []scanner.Option {
 	}
 	allExcludedRuleIDs = mergeWithoutDuplicates(allExcludedRuleIDs, tfsecConfig.ExcludedChecks)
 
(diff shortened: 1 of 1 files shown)


owenrumney (approved):

owenrumney:
Thanks for this @p9f

owenrumney (approved):

PR #1064, opened 2021-09-01
docker: specify the tags for the dockerfiles

chenrui333:
- would be good practice of specifying the base docker image tag number
- the base image tag dependencies would be managed by dependabot to ensure the latest and greatest


owenrumney (approved):
great idea, thanks @chenrui333

chenrui333:
My pleasure. :)

PR #1068, opened 2021-09-03
Use new rule identifiers in README examples

mathias-nyman:
The examples in the README still use the old legacy rule identifiers. This pull request changes the legacy rule IDs in the README to their corresponding new, more verbose, rule IDs.


owenrumney (approved):
thanks for this

PR #1080, opened 2021-09-14
Fix of two azurerm appservice rules

jan-mrm:
Hi, I am running into 2 findings of the following checks which I suppose are false positives:

- azure-appservice-detailed-error-messages-enabled
- azure-appservice-ftp-deployments-disabled

It seems to me that the checks themself are not checking the right thing. In both checks the property that is tested is not a direct property of the terraform azurerm resource but of a block within the resource.
I tried to provide this fix :)

"make pr-ready" does error but just because of the typos check for "iam" as far as I can see.


jan-mrm:
@owenrumney did you have a look by any chance? Should I change something? :)

owenrumney (approved):

owenrumney:
@jan-mrm - I butchered that on the conflict resolution :-( - apologies. Thanks for the contribution.

PR #1091, opened 2021-09-28
correcting azure app service recommendations

leesutcliffe:
This PR is to fix three issue within the `azurerm_app_service` resource

- detailed_error_messages_enabled
- ftps_state
- http_logs

### detailed_error_messages_enabled
`detailed_error_messages_enabled` should be moved within the `logs` child resource block as documented https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service#detailed_error_messages_enabled

### ftps_state
`ftps_state` should be moved to the `site_config` child resource block as documented 
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service#ftps_state

### http_logs
'http_logs' should be contain a `file_system` or `azure_blob_storage` child resource block. For simplicity, the purpose of the 'Good Example' code refers only to `file_system`. Additionally, the examples values for `retention_in_days` and `retention_in_mb` have been modified to values acceptable by Terraform
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service#file_system


internal/app/tfsec/rules/azure/appservice/http_logs_enabled_rule.go
```diff
@@ -64,8 +66,12 @@ resource "azurerm_app_service" "good_example" {
 			if logProps.MissingChild("http_logs") {
 				set.AddResult().
 					WithDescription("Resource '%s' does not have logs.http_logs enabled", resourceBlock.FullName()).WithBlock(logProps)
+				return
+			}
+			if logProps.MissingNestedChild("http_logs.file_system") {
```
owenrumney, line 71: Should this be a check for both types of http_logs? I guess the description would need changing too if you agree
```suggestion
			if logProps.MissingNestedChild("http_logs.file_system") && logProps.MissingNestedChild("http_logs.azure_blob_storage") {
```

owenrumney (changes requested):
This is great, thanks @leesutcliffe  - just one question, it might not be correct so just say

leesutcliffe:
Thanks for the suggestion, @owenrumney I have added two good examples and updated the logic.

owenrumney (approved):
Awesome, thanks @leesutcliffe

PR #1092, opened 2021-09-30
Change links

kmeenan-indx:
The packages were not building because the links were wrong - I just changed them from aquasecurity/... to IntelligentDX-LLC/...


owenrumney:
Hey @kmeenan-indx - guess you decided against this PR, changing the paths shouldn't be needed. Shout if you need assistance

PR #1094, opened 2021-10-01
add check for github vulnerability_alerts

iwarapter:


owenrumney (approved):
Thanks for this @iwarapter - much appreciated

PR #1095, opened 2021-10-01
Mysql Threat detection and required email security center check

brandon-bailey:
Adding a few more Azure terraform checks


internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_email_rule.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/aquasecurity/tfsec/internal/app/tfsec/block"
+
+	"github.com/aquasecurity/tfsec/pkg/rule"
+
+	"github.com/aquasecurity/tfsec/internal/app/tfsec/scanner"
+)
+
+func init() {
+	scanner.RegisterCheckRule(rule.Rule{
```
owenrumney, line 18: Hey @brandon-bailey  - we should just enrich https://github.com/aquasecurity/tfsec/blob/master/internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_rule.go with email - its essentially the same check

owenrumney (changes requested):

internal/app/tfsec/rules/azure/securitycenter/set_required_contact_details_email_rule.go
brandon-bailey: That makes sense, I was just hesitant to modify existing rules. I have removed this check in this PR and will brainstorm the best way to include it in the existing `set_required_contact_details_rule`

owenrumney: I think just chuck it in below the phone one and use the tests you'd already created will be enough.

so result for no phone, result for no email or both results. that should suffice

owenrumney (approved):

PR #1123, opened 2021-10-27
Update docs to have PATH info, strip some trailing whitespace

ahonnecke:
Signed-off-by: Ashton Von Honnecke <ashton@pixelstub.com>


README.md
```diff
@@ -49,11 +49,18 @@ Alternatively, install with Go:
 go install github.com/aquasecurity/tfsec/cmd/tfsec@latest
 ```
 
+Using newer versions of Go (1.11+) will place the tfsec binary in your go installation
+directory (`$HOME/go`).  If that has not been added to your $PATH, you will need to add that step to your `.bashrc`
+
+```bash
+export PATH=$PATH:~/go/bin
```
owenrumney, line 56:
```suggestion
export GOPATH=~/go
export PATH=~/$GOPATH/bin:$PATH
```

I think if we're going to guide people to put GOPATH onto their PATH we should do it in the accepted way

owenrumney:
closing due to inactivity

PR #1134, opened 2021-11-04
Rename to aws-elastic-search-encrypt-replication-group

holidayworking:
`aws-elastic-search-encrypt-replication-group` is a rule for Elasticache. Therefore, I rename to aws-elastic-search-encrypt-replication-group.


owenrumney (approved):
Good spot thanks

internal/app/tfsec/rules/aws/elasticache/encrypt_replication_group_rule.go
```diff
@@ -1,4 +1,4 @@
-package elasticsearch
+package elasticache
```
owenrumney, line 1: You need to update `Service: ` at line 20. Can't suggest it as it's not part of the PR :-(

owenrumney:
Thanks for this @holidayworking  - this has been resolved as part of another fix

PR #1165, opened 2021-11-30
Use KMS ARN in the ECR encryption good example

samm-git:
[Good example](https://github.com/aquasecurity/tfsec/blob/master/internal/app/tfsec/rules/aws/ecr/repository_customer_key_rule.go#L53) of the ECR rule wrongly suggest to use KMS `key_id` instead of `arn` object.

From the [documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecr_repository#encryption_configuration):

> encryption_configuration

>    encryption_type - (Optional) The encryption type to use for the repository. Valid values are AES256 or KMS. Defaults to AES256.
    kms_key - (Optional) The ARN of the KMS key to use when encryption_type is KMS. If not specified, uses the default AWS managed key for ECR.

As result if we using code from "good" example terraform would always detect false drift, as key_id is only part of ARN.

### DIFF
--- a/internal/app/tfsec/rules/aws/ecr/repository_customer_key_rule.go
+++ b/internal/app/tfsec/rules/aws/ecr/repository_customer_key_rule.go
@@ -52,7 +52,7 @@ resource "aws_ecr_repository" "good_example" {
 
 	encryption_configuration {
 		encryption_type = "KMS"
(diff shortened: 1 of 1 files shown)


owenrumney (approved):
Looks good. Thanks for taking the time to correct this.

Issue #1192, opened 2021-12-17
tfsec does not support terraform 1.1.0+ moved block


Cajga:
**Describe the bug**
We started to use the new feature of terraform 1.1.0+ called the moved block:
https://www.terraform.io/language/modules/develop/refactoring

Unfortunately, this results an error in tfsec:
```bash
$ tfsec --version
0.63.1
$ echo ${TFSEC_OPTS}
$ tfsec ${TFSEC_OPTS} .
/tmp/test.tf:27,1-6: Unsupported block type; Blocks of type "moved" are not expected here., and 1 other diagnostic(s)
Error: non zero exit code: 1: OCI runtime error
ERROR: Job failed: exit status 1
```

**To Reproduce**
Include a moved block into ferraform config and call tfsec.

**Expected behavior**
Would be nice if tfsec would support this block.

PR #1204, opened 2021-12-25
add support for moved blocks tf 1.1.0+

martijnvdp:
Fix to add support for the new moved blocks
see https://github.com/aquasecurity/tfsec/issues/1192


owenrumney (approved):
Thanks for doing this

liamg:
Thanks, we're anticipating a release in the next week, roughly.

Issue #1188, opened 2021-12-14
TFsec doesn't recognize terraform v1.1.0 modules after init


wilkej:
**Describe the bug**
We've got terraform and tfsec running in a gitlab pipeline. Recently I updated to TF 1.1.0 image and now TF sec shows the error

```hcl 
WARNING: Failed to load module: missing module with source 'terraform-aws-modules/s3-bucket/aws' -  try to 'terraform init' first
WARNING: Failed to load module: missing module with source 'terraform-aws-modules/eks/aws' -  try to 'terraform init' first
```
If I change the TF image in the pipeline to 1.0.11 it works. 

**To Reproduce**

1. run terraform 1.1.0 init
2. run tfsec and you'll see the errors

1. delete .terraform folder
2. run terraform 1.0.11 init
3. run tfsec you'll not see the errors

**Expected behavior**
TFsec should find the init modules for TF 1.1.0 

**Desktop (please complete the following information):**
 - OS: Ubuntu Linux

fgm:
Error also happens with default modules like `hashicorp/subnets`

martijnvdp:
looks like the source changed from 1.1.0 > example:
before : `{"Key":"endpoint_sg","Source":"**terraform-aws-modules/security-group/aws**","Version":"4.2.0","Dir":".terraform/modules/endpoint_sg"}`
now
`{"Key":"endpoint_sg","Source":"**registry.terraform.io/terraform-aws-modules/security-group/aws**","Version":"4.2.0","Dir":".terraform/modules/endpoint_sg"}`

martijnvdp:
hashicorps resolves to registry.terraform.com see  also : https://github.com/hashicorp/terraform/blob/main/internal/addrs/module_source_test.go
ive made a pr to change the check from source to module key name , but probably there other options/solutions 
also looked at go-getter but that uses a lot of dependencys

sergei-ivanov:
It is probably related to this major change in TF 1.1.x: https://github.com/hashicorp/terraform/pull/28854

sergei-ivanov:
@liamg  Can this be closed, now that #1206 has been merged?
And together with the fix for #1192, can we say that the latest RC versions are fully compatible with TF 1.1?
On my side, I put TF 1.1 upgrade on hold, waiting for TFSec to add first class support. Do you need help testing the RC, or is it not ready for early testing yet?

liamg:
[...]

PR #1206, opened 2021-12-26
fix tf v1.1.0 init module sources

martijnvdp:
This PR fixes the issue with the terraform v1.1.0+ module sources
by changing matching on module source to matching on module name with module key in the json metadata file.
module source is since tf 1.1.0 no longer always equal to the source in the metadata json file(.terraform\modules\modules.json) ,for instance terraform/hashicorp auto resolves to registry.terraform.com, googleapis.com to gcs etc..
see issue #1188

### DIFF
--- a/internal/app/tfsec/parser/load_module.go
+++ b/internal/app/tfsec/parser/load_module.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


liamg (approved):

liamg:
Thanks for this, we should have a release out in roughly a week.

PR #1220, opened 2021-12-28
DigitalOcean SSH keys check

atombrella:
There was a typo in the test code, and an unnamed test which fails with my changes. I'm not really sure how to fix it. I can work a bit more with the `digitalocean` checks.

`empty data for ssh key fails` is the problematic one. Help on how to fix is appreciated :100:

### DIFF
--- a/internal/app/tfsec/adapter/digitalocean/compute/adapt.go
+++ b/internal/app/tfsec/adapter/digitalocean/compute/adapt.go
@@ -6,5 +6,26 @@ import (
 )
 
 func Adapt(modules []block.Module) compute.Compute {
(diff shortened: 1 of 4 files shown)


owenrumney:
@atombrella  - We needed richer processing in one of the helper functions to handle getting the strings from the slice better. 

I've updated this branch with the required changes, let me know when you're happy with the change and I'll merge. 

Thanks for the contribution!

owenrumney (approved):

atombrella:
@owenrumney I'm happy. I think the slice/strings change is needed in more of the DigitalOcean adapters.

Issue #1269, opened 2022-01-08
AWS Autoscaling - convert tfsec checks to use defsec logic


vanesasejdiu:
Adapt AWS autoscaling launch configurations to Go structs based on #1194 instructions:

Remove tfsec logic from `internal/app/tfsec/rules/aws/autoscaling/enable_at_rest_encryption_rule.go`
`internal/app/tfsec/rules/aws/autoscaling/no_public_ip_rule.go`
`internal/app/tfsec/rules/aws/autoscaling/no_sensitive_info_rule.go`

Create adapt functions to populate the Go structs in `internal/app/tfsec/adapter/aws/autoscaling/adapt.go`

PR #1270, opened 2022-01-08
AWS Autoscaling - convert to use defsec

vanesasejdiu:
Closes: https://github.com/aquasecurity/tfsec/issues/1269

### DIFF
--- a/internal/app/tfsec/adapter/aws/autoscaling/adapt.go
+++ b/internal/app/tfsec/adapter/aws/autoscaling/adapt.go
@@ -1,10 +1,110 @@
 package autoscaling
 
 import (
+	"encoding/base64"
+
 	"github.com/aquasecurity/defsec/provider/aws/autoscaling"
+	"github.com/aquasecurity/defsec/types"
 	"github.com/aquasecurity/tfsec/internal/app/tfsec/block"
 )
 
 func Adapt(modules []block.Module) autoscaling.Autoscaling {
-	return autoscaling.Autoscaling{}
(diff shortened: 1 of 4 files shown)


internal/app/tfsec/adapter/aws/autoscaling/adapt.go
```diff
@@ -1,10 +1,110 @@
...
+
+	userDataVal := types.String("", *resource.GetMetadata())
+	if resource.GetAttribute("user_data").IsNotNil() {
+		userDataAttr := resource.GetAttribute("user_data")
+		userDataVal = userDataAttr.AsStringValueOrDefault("", resource)
+	} else if resource.GetAttribute("user_data_base64").IsNotNil() && resource.GetAttribute("user_data_base64").IsString() {
+		userDataBase64Attr := resource.GetAttribute("user_data_base64")
+		encoded, err := base64.StdEncoding.DecodeString(userDataBase64Attr.Value().AsString())
+		if err == nil {
+			userDataVal = types.String(string(encoded), *resource.GetMetadata())
```
liamg, line 98: Here we could use the metadata from `userDataBase64Attr` to highlight this is the source of the problem

liamg (changes requested):

internal/app/tfsec/adapter/aws/autoscaling/adapt.go
vanesasejdiu: @liamg Good catch! Updated

liamg (approved):

Issue #1337, opened 2022-01-14
Add "variables" to the custom check interface


gabrielleecredera:
**Is your feature request related to a problem? Please describe.** + **Describe the solution you'd like**
this would allow custom checks like the following example to be created
```json
"requiredTypes": ["resource"],
"requiredLabels": ["aws_s3_bucket"],
"matchSpec": {
    "action": "and",
    "predicateMatchSpec": [
        {
            "name": "bucket",
            "action": "isPresent",
            "assignVariable": "TFSEC_VAR_BUCKET_NAME"
        },
        {
            "name": "lifecycle_rule",
            "action": "isPresent",
            "subMatch": {
                "name": "id",
                "action": "startsWith",
                "assignVariable": "TFSEC_VAR_BUCKET_NAME"
            }
        }
    ]
}
```
in the above example, the bucket name will be stored in memory as "TFSEC_VAR_BUCKET_NAME" and can be retrieved by following checks.

PR #1338, opened 2022-01-14
add optional "assignVariable" attribute to "matchSpec"

gabrielleecredera:
- take "name" value and store in map with index "assignVariable" value
- e.g. you can check that an attribute exists, and assign the attribute value to a variable
- variables must start with "TFSEC_VAR_"
- the variable can then be used in any following matchSpec "value" attributes
- added the above to the custom check docs

solves #1337

### DIFF
--- a/internal/app/tfsec/custom/custom_context.go
+++ b/internal/app/tfsec/custom/custom_context.go
@@ -1,0 +1,31 @@
+package custom
+
+import "github.com/aquasecurity/tfsec/internal/app/tfsec/block"
+
+type customCheckVariables map[string]string
+
+type customContext struct {
+	module    block.Module
+	variables customCheckVariables
+}
+
+func NewEmptyCustomContext() *customContext {
+	return &customContext{
+		module:    nil,
+		variables: make(customCheckVariables),
+	}
+}
+
+func NewCustomContext(module block.Module) *customContext {
+	return &customContext{
+		module:    module,
+		variables: make(customCheckVariables),
+	}
+}
+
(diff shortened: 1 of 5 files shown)


gabrielleecredera:
(not sure how to address the cyclomatic complexity issue, mainly caused by the switch)

owenrumney (approved):

owenrumney:
> (not sure how to address the cyclomatic complexity issue, mainly caused by the switch)

I'll take a look at this shortly - thanks for the contribution :+1:

internal/app/tfsec/custom/custom_context.go
```diff
@@ -0,0 +1,31 @@
...
+
+type customContext struct {
+	module    block.Module
+	variables customCheckVariables
+}
+
+func NewEmptyCustomContext() *customContext {
+	return &customContext{
+		module:    nil,
+		variables: make(map[string]string),
```
liamg, line 15: Nitpicking but should probably be `make(customCheckVariables)`

internal/app/tfsec/custom/custom_context.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+}
+
+func NewCustomContext(module block.Module) *customContext {
+	return &customContext{
+		module:    module,
+		variables: make(map[string]string),
```
liamg, line 22: Nitpicking but should probably be `make(customCheckVariables)`

liamg (changes requested):
Looks good, one minor nitpick

liamg (approved):

Issue #1339, opened 2022-01-14
custom check: stop requiresPresence from ignoring subMatches


gabrielleecredera:
**Is your feature request related to a problem? Please describe.**
currently, you can't perform any additional checks on the resources found by requiresPresence.

**Describe the solution you'd like**
allow using subMatches the match all resources found by requiresPresence.

PR #1340, opened 2022-01-14
custom checks: requiresPresence's subMatches now get processed too

gabrielleecredera:
- did the above
- passing test case already exists, added fail test case

solves #1339


owenrumney (approved):

owenrumney (approved):
Looks good, thanks @gabrielleecredera

PR #1410, opened 2022-01-27
Add support for grouping of results

### LINKED ISSUE #1403: Group multiple instances of a module-based result
When an issue is detected inside a module, and the module is used multiple times in a parent or root module, the issue will be detected once for each use of the module.

It would be a better experience if all of these "duplicate" results were grouped, and additionally contained details of the usage locations of each module.

Raised in #1390

liamg:
Resolves #1403 

image


owenrumney (approved):
:rocket:

Issue #1415, opened 2022-01-27
Provide a feature to validate your custom checks work as expected


drewmullen:
**Is your feature request related to a problem? Please describe.**

Provide a feature to validate your custom checks work as expected. If we have custom checks in a central location, it would be good to have a way to validate those checks work as we expect.

**Describe the solution you'd like**
we provide HCL that passes and HCL that fails, `tfsec` would have a `test` cmd that would validate passing tests pass and failing tests fail

owenrumney:
we have a tool that validates the config file in the form of `./cmd/tfsec-checkgen validate <custom_check_file>` which - it doesn't check against HCL files but validates the contents of the check file itself.

I'll leave this open as enhancement for something to look at adding to `tfec-checkgen`

PR #1436, opened 2022-01-30
feat: add tfsec-checkgen test-check command to test custom check against tf test code

gabrielleecredera:
- new command `tfsec-checktest test-check <custom_check_file> -p <pass.tf> -f <fail.tf>`
- number of `--pass`/`-p`/`--fail`/`-f` flags are optional and variable
- had to modify the internal custom package to expose functions
- subcommand `test-check` used to prep for future `test-checks` command that runs tests against more than one custom check files

partially solves #1415


owenrumney:
Could this be added as an additional command to tfsec-checkgen?

gabrielleecredera:
I added it as a new command as I think I've seen somewhere that the original purpose of checkgen is to generate custom checks? I'm not sure if that vision is still true tho, let me know if adding it into checkgen is preferred. :)

owenrumney:
> I added it as a new command as I think I've seen somewhere that the original purpose of checkgen is to generate custom checks? I'm not sure if that vision is still true tho, let me know if adding it into checkgen is preferred. :) (unrelated: if the vision is still true, what's the solution expected to look like? some cli q&a to generate a custom check file?)

Yeah, lets stick with it being part of checkgen since its all the same sort of thing.

In answer to the other question, before we moved to defsec we had tfsec-skeleton which asked you a number of questions about services, required labels, required resources. I think it would do similar to that.... I think there is only so far we can go with automating the generation of custom checks so it will be a bit trial and error

owenrumney (approved):
Thanks again @gabrielleecredera  - your contributions are really appreciated :heart:

owenrumney:
hey @gabrielleecredera  - looks like a goimports issue breaking the build

owenrumney (approved):

Issue #1437, opened 2022-01-30
feat: add subMatchOne attribute to custom checks checking for one match only


gabrielleecredera:
**Is your feature request related to a problem? Please describe.**
Currently, the subMatch attribute allows checking for all matching "sub-blocks", sometimes it's helpful to check if there's only 1 match in all "sub-blocks". For e.g., we may want to check that a `aws_dynamodb_table` is at least replicated in "us-east-2" (other replications can be non "us-east-2"), this check is currently impossible to implement with custom checks:
```tf
resource "aws_dynamodb_table" "example" {
  name             = "example"
  hash_key         = "TestTableHashKey"

  attribute {
    name = "TestTableHashKey"
    type = "S"
  }

  replica {
    region_name = "us-east-2"
  }

  replica {
    region_name = "eu-west-2"
  }
}
```

**Describe the solution you'd like**
`subMatchOne` attribute added to custom checks allowing the above check

**Describe alternatives you've considered**
explored using `subMatch` with a combination of the `not`/`or`/`and` actions to perform the above check

PR #1438, opened 2022-01-30
feat: add subMatchOne attribute to custom checks checking for one match only

gabrielleecredera:
- added subMatchOne attribute
- added tests, and docs

solves #1437


gabrielleecredera:
(linting will fail, as evalMatchSpec complexity went from 15(max) to 16, unsure how to fix tho, could consider collapsing the switch, creating a `matchFunctions` similar object for the instant return cases)

owenrumney (approved):

Issue #1472, opened 2022-02-01
feat: Include preConditions in Documentation


arkaprava-jana:
Include preConditions in Documentation

owenrumney:
Resolved by #1473

PR #1473, opened 2022-02-01
docs: document preConditions in custom checks

gabrielleecredera:
solves #1472

### DIFF
--- a/internal/pkg/custom/custom_check.go
+++ b/internal/pkg/custom/custom_check.go
@@ -107,7 +107,7 @@ type MatchSpec struct {
 	Name               string      `json:"name,omitempty" yaml:"name,omitempty"`
 	MatchValue         interface{} `json:"value,omitempty" yaml:"value,omitempty"`
 	Action             CheckAction `json:"action,omitempty" yaml:"action,omitempty"`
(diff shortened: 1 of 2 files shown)


internal/pkg/custom/custom_check.go
```diff
@@ -107,7 +107,7 @@ type MatchSpec struct {
 	Name               string      `json:"name,omitempty" yaml:"name,omitempty"`
 	MatchValue         interface{} `json:"value,omitempty" yaml:"value,omitempty"`
 	Action             CheckAction `json:"action,omitempty" yaml:"action,omitempty"`
-	PreConditions      []MatchSpec `json:"preconditions,omitempty" yaml:"preconditions,omitempty"`
+	PreConditions      []MatchSpec `json:"preConditions,omitempty" yaml:"preConditions,omitempty"`
```
liamg, line 110: I think this change is going to break existing custom check configurations.

gabrielleecredera: I've tested it, apparently, the parser converts the keys to lowercase

liamg (approved):
LGTM :+1:

PR #1486, opened 2022-02-02
fix: updated rules.md with missing rule

danielsteman:
fix(rules): I needed to ignore this rule in my project to get an exception on one of my Terraform variables


owenrumney (approved):
Thanks @danielsteman

PR #1515, opened 2022-02-07
docs: aws-cloudfront-use-secure-tls-policy clarification

jrobison-sb:
- The good example for aws-cloudfront-use-secure-tls-policy wasn't a possible configuration, because it's not possible to set a custom TLS policy when `cloudfront_default_certificate = true`. This fixes the example and also adds a note that this rule is impossible to satisfy when `cloudfront_default_certificate = true` (eg. when using the cloudfront.net domain name).
- https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_ViewerCertificate.html#cloudfront-Type-ViewerCertificate-MinimumProtocolVersion ctrl-f see the last paragraph in the MinimumProtocolVersion section.


owenrumney:
Thanks @jrobison-sb  - I've fixed this at source (aquasecurity/defsec) so closing this. The docs are generated from the info in defsec!

PR #1525, opened 2022-02-10
WIP: chore: Switch to defsec/trivy-config-parsers for parsing and adaptation

liamg:


owenrumney (approved):
it looks like there should be some sort of flow chart on this?

PR #1539, opened 2022-02-17
fix: Really force all directories recursively when `--force-all-dirs` is used

liamg:


owenrumney (approved):
I thought you said this was redundant now?

liamg:
> I thought you said this was redundant now?

I keep forgetting to tell you about this.

PR #1552, opened 2022-02-19
fixed aws, azure, checks etc.. 404s to github.io doc sites

leewalter:
all checks ref to https://aquasecurity.github.io/tfsec/xxx are now 404, to fix it:
1/ either add dir listing as in https://ongclement.com/blog/github-pages-indexing-directory-copy
or 2/ ref them to github.com instead as in my fix here 
Any fix is better than all 404s now : )

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,16 +164,16 @@ There are also checks which are provider agnostic.
 
 | Checks |
 |:---|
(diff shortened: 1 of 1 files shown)


owenrumney:
Thanks for this @leewalter  - I've fixed this more permanently in #1553 so I'll close this one if you don't mind.

Thanks again

PR #1553, opened 2022-02-20
add index pages for providers and services

### LINKED ISSUE #1478: bug: documentation index pages are missing
**Describe the bug**
Index pages in the documentation aren't being created - https://aquasecurity.github.io/tfsec/latest/checks/aws/home/

**To Reproduce**
1. Go to the link -https://aquasecurity.github.io/tfsec/latest/checks/aws/home/
2. See that it doesn't work

**Expected behavior**
Expected the link to work - https://aquasecurity.github.io/tfsec/latest/checks/aws/home/

owenrumney:
Resolves #1478 and makes the changes in #1552 more repeatable


cmd/tfsec-docs/indexes.go
```diff
@@ -0,0 +1,102 @@
...
+const providerIndexPageTemplate = `---
+title: {{$.DisplayName}}
+---
+
+# {{$.DisplayName}}
+
+## Services
+
+{{range $link := $.Services}}
+- [{{.Name}}]({{.Path}})
```
liamg, line 84: This might be fine but not sure if the markdown will render with 2 newlines in between each point?

liamg (changes requested):

cmd/tfsec-docs/indexes.go
owenrumney: image
It renders as a single line

liamg (approved):

PR #1561, opened 2022-03-01
docs: fixed checks pages is 404

int-tt:
it was discussed in #1478 #1552, but the link of checks is still 404.
Since there is no `home/` , I thought it was correct and corrected it.

ex: https://aquasecurity.github.io/tfsec/latest/checks/aws/home/ -> https://aquasecurity.github.io/tfsec/latest/checks/aws/

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,16 +164,16 @@ There are also checks which are provider agnostic.
 
 | Checks |
 |:---|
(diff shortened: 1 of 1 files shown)


owenrumney (approved):
thanks for this @int-tt

PR #1571, opened 2022-03-04
chore: make Make more friendly for macs

reedloden:
Additionally:
* Fix typo in CONTRIBUTING
* Add `vendor` to .gitignore / codespell skip

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 IMAGE := tfsec/tfsec
-SHELL := /bin/bash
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,5 +1,4 @@
 IMAGE := tfsec/tfsec
-SHELL := /bin/bash
```
owenrumney, line 2: This has implications for the github action which we need to resolve before removing from here

PR #1701, opened 2022-04-27
docs: fix branch names in sarif github action

gmarmstrong:
Branch names were backwards: `uses: actions/checkout@main` should be `@master`, whereas `uses: tfsec/tfsec-sarif-action@master` should be `@main`.

Note that the former, `actions/checkout@main` _does_ exist, it's just no longer the active branch.


owenrumney (approved):

owenrumney:
good spot @gmarmstrong  - thanks

PR #1725, opened 2022-05-11
feat: Add customisable code themes

liamg:


internal/app/tfsec/cmd/output.go
```diff
@@ -85,7 +85,7 @@ func outputFormat(w io.Writer, addExtension bool, baseFilename, format, fsRoot,
 	switch strings.ToLower(format) {
 	case "lovely", "default":
 		alsoStdout = true
-		factory.WithCustomFormatterFunc(formatter.DefaultWithMetrics(metrics, conciseOutput))
+		factory.WithCustomFormatterFunc(formatter.DefaultWithMetrics(metrics, conciseOutput, codeTheme))
```
owenrumney, line 88: Do we need to validate input here? Or does defsec do it?

internal/pkg/formatter/default.go
```diff
@@ -293,9 +293,22 @@ func printMetadata(w io.Writer, result scan.Result, links []string, isRego bool)
...
+		scan.OptionCodeWithTruncation(true),
+	}
+
+	switch theme {
+	case "dark":
+		codeOpts = append(codeOpts, scan.OptionCodeWithDarkTheme())
+	case "light":
+		codeOpts = append(codeOpts, scan.OptionCodeWithLightTheme())
+	default:
+		codeOpts = append(codeOpts, scan.OptionCodeWithTheme(theme))
```
owenrumney, line 308: ignore previous question

owenrumney: but what if I write junk in here - does it fallback in defsec?

liamg: Yeah it falls back to a default dark one.

liamg: Thought i'd leave it undocumented so we can play with it, but tell people to only use dark/light

owenrumney (approved):

PR #1772, opened 2022-06-08
feat: Install git on the CI Docker

### LINKED ISSUE #1685: feat: preinstall git in tfsec-ci container
I am trying to use the `tfsec-ci` container in CircleCI to scan our terraform code. To do this, it needs `git` or `ssh` present, and also needs to be able to write to the filesystem. Instead, because the user is nobody, the directory is not writable, and because there's no sudo, it's not possible to run, e.g., `apk update && apk add git` or something

Let me know if there's a different way to do this or a better example - would be happy to follow that.

* To have a generic (non-root) user and home directory in the `tfsec-ci` container that's writable by the user the container runs as
* To have, or be able to install, `git` and / or `ssh` (though if #1 is solved, I think Circle will be able to bootstrap it via another method)

edit: I guess this is one other way to do it:
https://github.com/mycodeself/tfsec-orb/blob/ad51b17dad375aa70aac95a3ab93c2b6da395539/src/executors/default.yml#L5

* Use the hack above to change to a different username
* Use a different base container and bootstrap tfsec via some other method
* Use a different scanning tool

**Additional context**
image

owenrumney:
resolves #1685 

Signed-off-by: Owen Rumney <owen.rumney@aquasec.com>


wyardley:
Thanks 🎉 
Much appreciated!

PR #1782, opened 2022-06-13
feat: Add markdown and html output formats

### LINKED ISSUE #1729: feat: Markdown output

We want to push the output to our MR on GitLab. To have it nicely formatted we need to use Markdown. With a previous version, we could run a `sed` command to create headings. Since the output was changed, this no longer works. 

The proper solution would be Markdown output

As described, we used `sed` in the past, but the output isn't stable

**Additional context**

I'd be willing to implement but need a pointer on where to start / what the preferred way is. I can create a Markdown formatter or try to reuse the `text` or `default` formatter somehow.

liamg:
Resolves #1729


owenrumney (approved):
Looks good

PR #1839, opened 2022-07-27
feat: Add exclude-ignores flag and --config-file attribute

alexandrupopafc:
This is intended for adding exclude-ignores flag and --config-file attribute.

Update to github.com/aquasecurity/defsec v0.69.0
https://github.com/aquasecurity/defsec/pull/767

In some cases, we want to enforce some rules and remove the ability to #tfsec:ignore


liamg (approved):
LGTM, thank you! :rocket:

PR #1870, opened 2022-08-25
fix: Use `mktemp` instead of `/tmp` to securely manage temporary directories

reedloden:
To avoid TOCTOU and other security issues, use `mktemp` to create a download directory for tfsec and checksum. Remove all usage of `/tmp` entirely.


dawez:
very nice. tested on linux mint and works as expected. :+1:

liamg (approved):
LGTM, thanks! :+1:

PR #1958, opened 2022-12-01
docs: fix good example typo in `google-compute-enable-shielded-vm-vtpm`

patrickblackjr:


owenrumney:
thanks for this @patrickblackjr - this actually needs updating in `aquasecurity/defsec` I've raised it as a PR in there https://github.com/aquasecurity/defsec/pull/1129

Will close this once that is resolved.

Thanks again

Issue #1943, opened 2022-11-02
Add URI handler to use Github PAT to download `--config-file-url` for a private github repo


nitrocode:
I'd like to re-use the config file from a private repo. How can I do that with `--config-file-url` ? Is there a way I can pass in my PAT ?

```sh
export GITHUB_TOKEN=$(gh config get -h github.com oauth_token)
```

Is there a way to use an `ssh` url ?

https://superuser.com/a/1687399

Doing something like this would be ideal because then it would re-use my current ssh configuration instead of a PAT

```
tfsec --config-file-url ssh://git@github.com/foo/bar
```

owenrumney:
Hey @nitrocode - only http[s] is supported for config download https://github.com/aquasecurity/tfsec/blob/master/internal/app/tfsec/cmd/flags.go#L318 

So you could use something like 

```
tfsec --config-file-url https://raw.githubusercontent.com/aquasecurity/tfsec/master/_examples/with_config_overrides/.tfsec/config.yml
```

nitrocode:
Are you saying that it is not possible to download a remote config file from a private GitHub repo?

Perhaps then this question can serve as a feature request for an ssh handler.

nitrocode:
Let's make it unstale then :D

nitrocode:
Let's make it unstale then :D

nitrocode:
Unstale

simar7:
Our efforts have moved to Trivy. Please file a discussion in the Trivy repo if you see fit.

PR #1963, opened 2022-12-17
Set github stale bot message "days" to its config

nitrocode:
## what
- Set github stale bot message "days" to its config

## why
- It's confusing when the message says that the stale bot will close the issue/pr in X number of days and that X number of days is not the same as the days configured in the bot itself. This change should set the numbers to be the same.

## references
- https://github.com/aquasecurity/tfsec/issues/1943


owenrumney (approved):
thanks

Issue #1986, opened 2023-02-05
bug: install linux script sha256 checksum not working correctly 


aiell0:
**Describe the bug**
When trying to install tfsec using the install linux script, it fails with the following error:

```
=====================================
Looking up the latest version...
Downloading tfsec v1.28.1
Downloaded successfully
Validate checksum
sha256sum: unrecognixzed option `--ignore-missing`
Try `sha256sum --help` for more information.
Failed to verify checksum
```

Additionally, once this was fixed, the checksum is verified against the entire checksums txt file, which fails. There is a need to extract only the checksum you need and verify that that instead of the entire file.

**To Reproduce**
1. Run the command `curl -s https://raw.githubusercontent.com/aquasecurity/tfsec/master/scripts/install_linux.sh | bash` on a linux machine

**Expected behavior**
```
============================================
Looking up the latest version...
Downloading tfsec v.1.28.1
Downloading tfsec-linux-amd64...
Downloaded file "tfsec-linux-amd64" successfully
Downloading tfsec-checkgen-linux-amd64...
Downloaded "tfsec-checkgen-linux-amd64" successfully
Downloading tfsec_checksums.txt...
Downloaded file "tfsec_checksums.txt" successfully
Checksum verified successfully
```

**Output of your tfsec command with --debug flag**
N/A

**System Info**

 - tfsec version: `v1.28.1`
 - terraform version: `n/a`
 - OS: `linux/amd64 and darwin`

**Example Code**
n/a

**Additional context**
None

neilherbertuk:
I'm getting almost the same on the latest alpine, however, my error is`sha256sum: unrecognized option: quiet`

Nil

chb0github:
I would say that, as a matter of implementation of the install script, it should:
1. use sh and not bash, since with alpine you need to add bash
2. remove (or make optional) the checksum switch, so that it can be externalized. 

Alternative is to figure out which version of checksum you have., Which is different between mac and linux for sure, and then switch-case your way to the write command and args

aiell0:
Just for everyone's reference, https://github.com/aquasecurity/tfsec/pull/1987 was merged recently.

@chb0github I don't disagree with your suggestions, but felt that it was out of scope for the PR I had open. I don't have it in me to make the changes you're suggesting so will leave it up to someone else to tackle it :)

adamgerhartz:
> I'm getting almost the same on the latest alpine, however, my error is`sha256sum: unrecognized option: quiet`

[...]

PR #1987, opened 2023-02-05
fix: sha256 command

aiell0:
https://github.com/aquasecurity/tfsec/issues/1986

### DIFF
--- a/scripts/install_linux.sh
+++ b/scripts/install_linux.sh
@@ -15,22 +15,25 @@ arch=$(get_machine_arch)
 
 echo "arch=$arch"
 
-remote_filename="tfsec"
 local_filename="tfsec"
 case "$(uname -s)" in
   Darwin*)
-    remote_filename+="-darwin-${arch}"
+    remote_filename="$local_filename-darwin-${arch}"
+    checkgen_filename="$local_filename-checkgen-darwin-${arch}"
     ;;
   MINGW64*)
-    remote_filename+="-windows-${arch}"
+    remote_filename="$local_filename-windows-${arch}"
+    checkgen_filename+="$local_filename-checkgen-windows-${arch}"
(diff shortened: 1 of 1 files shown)


scripts/install_linux.sh
```diff
@@ -58,25 +62,25 @@ fi
...
-echo "Validate checksum"
-curl --fail --silent -L -o "${download_path}/${checksum_file}" "https://github.com/aquasecurity/tfsec/releases/download/${version}/${checksum_file}"
-checksum_dl_val=$?
-if [ $checksum_dl_val -ne 0 ]; then
-  echo "Failed to download checksum file ${checksum_file}"
-  exit $checksum_dl_val
-fi
+download_file() {
+  echo "Downloading $3..."
+  curl --fail --silent -L -o "$1/$3" "https://github.com/aquasecurity/tfsec/releases/download/$2/$3"
```
chb0github, line 67: recommendation:

```bash
download_file() {
   local download_path=${1:?Download path no supplied}   
   local version=${2:?No version supplied}
   local file=${3:?File to download not supplied}
   curl --fail-with-body --fail --silent -L -o "${download_path}/${file}" "https://github.com/aquasecurity/tfsec/releases/download/${version}/${file}"
}
```
This has a few advantages:
1. it asserts that params were passed or the script returns an error
2. make it obvious which params go where and represents what by naming them
3. On failure, outputs the body if there was an error (optional, but good for debug)

chb0github (commented):
you should use named variables for clarity and bash assertions

scripts/install_linux.sh
aiell0: @chb0github thank you for the detailed review. Will edit and resubmit. Thank you!

giorod3 (dismissed):

PR #2044, opened 2023-04-19
Add missing legacy code AWS097

georgepstaylor:
We found this code `AWS097` in an exclusion list which we are migrating to the new string based id codes.
In the legacy map I was able to find all of the codes apart from this one which seems to be missing.

Researching this (as most of the docs related to this legacy code are gone from the web), I believe I have found the corresponding new style ID: 

* https://aquasecurity.github.io/tfsec/v0.63.1/checks/aws/iam/block-kms-policy-wildcard/

### DIFF
--- a/internal/pkg/legacy/map.go
+++ b/internal/pkg/legacy/map.go
@@ -58,6 +58,7 @@ var IDs = map[string]string{
 	"AWS083": "aws-elb-drop-invalid-headers",
 	"AWS004": "aws-elb-http-not-used",
 	"AWS010": "aws-elb-use-secure-tls-policy",
+	"AWS097": "aws-iam-block-kms-policy-wildcard",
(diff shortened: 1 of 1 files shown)


AnaisUrlichs:
Hi @georgepstaylor -- thank you for sharing. Quick question, you are using tfsec in your org? We suggest folks to switch over to Trivy https://github.com/aquasecurity/tfsec/discussions/1994

georgepstaylor:
> Hi @georgepstaylor -- thank you for sharing. Quick question, you are using tfsec in your org? We suggest folks to switch over to Trivy #1994

Hey @AnaisUrlichs, at the moment there are still some places we use tfsec but will migrate over to trivy in the near future.

A lot of us weren't actually aware that tfsec had been deprecated until recently - could be worth putting a banner at the top of this repo's read me?

Issue #2066, opened 2023-06-13
feat: add migration notice to tfsec cli


simar7:
As discussed [here](https://github.com/aquasecurity/tfsec/pull/2020#issuecomment-1548767236), the next (and final release) of tfsec can have a callout for migration to Trivy, embedded as a preamble (similar to motd).

PR #2067, opened 2023-06-13
feat: Add a transition message

simar7:
Fixes: https://github.com/aquasecurity/tfsec/issues/2066

### DIFF
--- a/cmd/tfsec/main.go
+++ b/cmd/tfsec/main.go
@@ -8,7 +8,21 @@ import (
 	"github.com/aquasecurity/tfsec/internal/app/tfsec/cmd"
 )
 
+const transitionMsg = `
+======================================================
+tfsec is joining the Trivy family
+
+tfsec will continue to remain available 
+for the time being, although our engineering 
+attention will be directed at Trivy going forward.
+
+You can read more here: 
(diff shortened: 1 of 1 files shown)


itaysk (approved):

kawikao:
This is breaking the tfsec-sarif-action upload because it's not valid sarif.

cmd/tfsec/main.go
```diff
@@ -8,7 +8,21 @@ import (
...
+for the time being, although our engineering 
+attention will be directed at Trivy going forward.
+
+You can read more here: 
+https://github.com/aquasecurity/tfsec/discussions/1994
+======================================================
+`
+
 func main() {
+	fmt.Print(transitionMsg)
```
yu-iskw, line 25: @simar7 @itaysk Can we print it out as standard error? The standard out message disable us to parse the result of JSON format.

yu-iskw: I opened a pull request. Let's discuss that there. 

https://github.com/aquasecurity/tfsec/pull/2099

itaysk: you're right, sorry. It shouldn't be printed in machine-readable formats like json, csv, junit, sarif, etc.

PR #2162, opened 2024-12-12
CVE-2024-45337: Misuse of ServerConfig.PublicKeyCallback may cause authorization bypass

jdesouza:
CVE-2024-45337: Misuse of ServerConfig.PublicKeyCallback may cause authorization bypass


simar7 (approved):
Thanks.

PR #2177, opened 2025-05-12
CVE-2025-46569: Improper Control of Generation of Code ('Code Injecton')

jdesouza:
CVE-2025-46569: Improper Control of Generation of Code ('Code Injecton')
https://avd.aquasec.com/nvd/2025/cve-2025-46569/


simar7:
tfsec does not run OPA in server mode.

jdesouza:
> tfsec does not run OPA in server mode.

Cool, thanks for your replay.
Anyways the CVE keep been caught by Trivy scanning and customers are always questioning/concerning about Critical or High CVEs.

jdesouza:
@simar7 - I would move to trivy but last year when we tried to move to trivy to replace tfsec we missed some features:
- validating custom policies against terraform plan and not to the original files
- we would need to use Rego instead of current tfsec approach for creating custom checks
Are those statements still valid? I mea,n is there a way to move to trivy using old tfsec custom checks against original terraform files and not plans?

simar7:
> @simar7 - I would move to trivy but last year when we tried to move to trivy to replace tfsec we missed some features:
> 
> * validating custom policies against terraform plan and not to the original files

Do you have an example? Trivy is able to scan terraform plan as well.  https://trivy.dev/latest/docs/coverage/iac/terraform/

> * we would need to use Rego instead of current tfsec approach for creating custom checks

Yes we no longer have support for checks written in Go. All custom checks must be written in Rego. But if this is an issue, I would like to understand what is the friction point. We have written some docs on writing custom checks https://trivy.dev/latest/tutorials/misconfiguration/custom-checks/ but if they are not enough or if you are running into other issues please let us know.

jdesouza:
> > @simar7 - I would move to trivy but last year when we tried to move to trivy to replace tfsec we missed some features:
> > 
> > * validating custom policies against terraform plan and not to the original files
> 
> Do you have an example? Trivy is able to scan terraform plan as well. https://trivy.dev/latest/docs/coverage/iac/terraform/
> 
> > * we would need to use Rego instead of current tfsec approach for creating custom checks
> 
> Yes we no longer have support for checks written in Go. All custom checks must be written in Rego. But if this is an issue, I would like to understand what is the friction point. We have written some docs on writing custom checks https://trivy.dev/latest/tutorials/misconfiguration/custom-checks/ but if they are not enough or if you are running into other issues please let us know.


- It's the opposite, we scan terraform files with tfsec in a repo but looks like trivy scans only the plan right?
- The point of friction is migrating the existing checks to Rego; also some people don't love Rego

simar7:
As mentioned In the docs Trivy can scan both.

